PR #102, opened 2022-11-22
Update readme 98

AzaniaBG:


AzaniaBG:
> @AzaniaBG Great Job!
> 
> (I actually have some suggestions, but I think we can go ahead and merge it first and make other PRs later.)

Sounds good, @yoyoyojoe ~ probably best to get something up and then refine it from there :) Maybe we can add to the next meeting's Agenda?

yoyoyojoe (approved):
Looks good. We can make incremental and minor changes later, and we can discuss what is the preferred process for making these changes as we move forward.

PR #122, opened 2022-12-21
Sync last reverted PR.

yoyoyojoe:
Seems like last PR was not merge properly. Here's another attempt.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "cSpell.words": [
+    "VRMS"
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # PeopleDepot
 
-PeopleDepot is a project of Hack for LA, a brigade of Code for America. PeopleDepot aims to provide a single source of truth as the backend infrastructure and database store for all Hack for LA projects, including data about people, program areas, and projects. PeopleDepot uses Django for its database and REST Framework and serves as a repository of information for other infrastructure projects (e.g., VRMS, Hack for LA Website, Civic Tech Index, Tables).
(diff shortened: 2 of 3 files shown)


yoyoyojoe:
Fixed typos, shifted heading levels, changed some words, made last section complete.

ethanstrominger (changes requested):
Spelling in settings.json - very cool.  I thought we had agreed that top level headings (1., 2., etc) should be capitalized.  IMO looks better to have them capitalized.  I see you also indented all headings except Overview.  This implies that everything is a subsection of Overview.  Either the title should be changed to “Contributing to People Depot” or heading levels restored.

yoyoyojoe:
Thanks @ethanstrominger for your feedback! Yes, your observations are correct and the changes are intentional, but I agree with you that "Overview" would no longer be the best title. 

1. Heading capitalization: 
You're right that we have agreed on capitalizing all 1st character of all words in heading level 1, and Sentence casing for heading level 2 & lower. At least that's what I remembered. 

2. The indentation was intentional: 
Reading through the doc, something made me feel that all the prior heading level 1's were more of a subordinate to "Contributing to People Depot", therefore the reasoning to make most of them heading level 2, and placing them under "Overview". What do you think about having only the 2 following headings at level 1? I think by doing this, it makes more sense and we're also sticking to point 1 regarding heading capitalization.

- Contributing to People Depot
- Appendix

3. By the way, if you click on the more button to the right of file changed `...` and go to `View File`

image

You'll be able to see the file with the changes, and you can see the Table of Contents on the upper left corner there. It'd be too many headings for me to type it here.

image

image

yoyoyojoe:
What it looks like now. 

image 
image

ethanstrominger (approved):

PR #125, opened 2023-01-05
Missing recurring event test

fyliu:
I could find the original version but not how it got erased. It must have happened during a bad merge somewhere.


yoyoyojoe (approved):
LGTM

PR #134, opened 2023-02-19
Feature/add location model

### LINKED ISSUE #44: Create Table: location
### Overview

We need to create the location table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items

- [ ] add issue link to [the tables tab of the spreadsheet](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit#gid=1228509670)
- [ ] create a single model in Django (defining schema)
- [ ] write a test for the relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project).
- [ ] write an API end point
- [ ] write API unit tests
- [ ] document the endpoint

### Resources/Instructions

  - ERD
  - Table and Field Definitions
  - API Endpoint

### Description

Location to be used in event and recurring event.

### Data Fields

   - [x] name - varchar
   - [x] address_line1 - varchar
   - [x] address_line2 - varchar
   - [x] city - varchar
   - [x] state - varchar
   - [x] zip - varchar
   - [x] phone - varchar

   - None

### Associated Tables

   - [x] event (one-to-many)
   - [x] event_history (one-to-many)
   - [x] recurring_event (one-to-many)

   - project (one-to-many) #16

chelseybeck:
fixes #44 

These changes add the location table model as well as the API endpoints and tests


chelseybeck:
@fyliu please review at your convenience (no rush on my end). I went ahead and took up the issue to build this table/model since the event table is dependent on it. Really I just wanted to get into working w/ Django and models and the easiest way was to build :)

fyliu:
I've been resolving automation issues on my end. I have essentially a new computer after my old ssd suddenly stopped working last month.

But before that, I had set up pre-commit to run some checks. Unfortunately, with the choice we made in the project structure, it's not possible to install pre-commit in requirements.txt. So I have it installed globally, which I don't like. I did set up a local pyenv version in the project root and set it to 3.10.0, which is the same as the project, so it installed under that version. I want to figure out a better way when there's time. I think it can run in a docker container.

[This](https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories) is the recommended setup for it and will be convenient if you will work with other pre-commit projects. If not, then `pre-commit install` should work in the project root.

I had to fix a few pre-commit configs
- fix flake8 repo address
- update isort to 5.12 to resolve some format issues with my newer version of poetry
- update hadolint to use their docker setup so we don't have to install it on the local system

Now pre-commit should work. You can run it using `pre-commit run --all-files`

There are several things in the PR:
1. isort auto-arranged some import statements
2. flake8 complained about zip being sort of a reserved word
3. pre-commit auto-formatted some whitespace issues
4. black did some auto-formatting
5. The migration files can be combined into a single one for the PR

The isort, flake8, black items are easier to resolve by running `./scripts/lint.sh` from the project root.

To roll back migrations, delete them, and have the ORM regenerate a single file, you can run these commands from project root:
```bash
docker-compose exec web python manage.py migrate core 0008
rm app/core/migrations/{0009*,0010*,0011*,0012*,0013*}
./scripts/migrate.sh
```

I notice one thing about the actual code, but I need to look through that more to comment on it.

chelseybeck:
Ah, that all makes sense. I did have issues with precommit throwing an error related to Flake 8, so I added the no-verify flag in order to push the commits. I'll make those changes and see if I'm able to commit without that flag. Thanks @fyliu

fyliu:
@chelseybeck it takes a new computer to figure out these config issues. I'd be better to dockerize it.

I have the changes in my [branch/fork](https://github.com/fyliu/peopledepot/tree/add_location_model). You're welcome to cherry-pick them into your branch, especially the first few that fixes the pre-commit setup. The other ones you can try doing yourself. Or just use them if this first issue is too much to take in.

app/core/api/urls.py
```diff
@@ -16,6 +17,7 @@
 router.register(r"sponsor-partners", SponsorPartnerViewSet, basename="sponsor-partner")
 router.register(r"faqs", FaqViewSet, basename="faq")
 router.register(r"faqs-viewed", FaqViewedViewSet, basename="faq-viewed")
+router.register(r"location", LocationViewSet, basename="location")
```
fyliu, line 20: The first argument should be "locations". It's what's used for the endpoint prefix like /api/locations/

app/core/models.py
```diff
@@ -215,3 +215,19 @@ class Meta:
...
+class Location(AbstractBaseModel):
+    """
+    Location for event
+    """
+    name = models.CharField(max_length=255, unique=True, verbose_name='Location name')
+    address_line_1 = models.CharField(max_length=255, unique=False)
+    address_line_2 = models.CharField(max_length=255, unique=False)
+    city = models.CharField(max_length=100, unique=False)
+    state = models.CharField(max_length=2, unique=False)
+    zip = models.CharField(max_length=10, unique=False)
```
fyliu, line 229: The linter didn't like this one because it's overriding a python builtin. You can simply change it to zipcode. And maybe add more code to make it look like "zip" to the outside if that's what it should be.

app/core/models.py
```diff
@@ -215,3 +215,19 @@ class Meta:
...
         return f"{self.faq} viewed at {self.created_at.strftime('%b %d %Y %H:%M:%S')}"
+
+
+class Location(AbstractBaseModel):
+    """
+    Location for event
+    """
+    name = models.CharField(max_length=255, unique=True, verbose_name='Location name')
+    address_line_1 = models.CharField(max_length=255, unique=False)
+    address_line_2 = models.CharField(max_length=255, unique=False)
```
fyliu, line 226: address_line_2 needs blank=True to make it optional

app/core/models.py
```diff
@@ -215,3 +215,19 @@ class Meta:
...
+    city = models.CharField(max_length=100, unique=False)
+    state = models.CharField(max_length=2, unique=False)
+    zip = models.CharField(max_length=10, unique=False)
+    phone = models.CharField(max_length=15, unique=False, verbose_name='Phone number')
```
fyliu, line 230: Should use PhoneNumberField that's used above for another model

fyliu (changes requested):
Thanks for working on this.

Just a few specific things. Other things can be corrected by the linters.

chelseybeck:
Changes were made in a new branch w/ a fresh install. Refer to [this pr](https://github.com/hackforla/peopledepot/pull/135) for final code.

Issue #54, opened 2022-09-08
Create Table: language :smile: 


ericvennemeyer:
### Overview

We need to create the language table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

A table and a model are the same thing

### Action Items

- [x] identify and document table description (see spreadsheet under Resources)
  - [x] if not, reach out to PD leads
- [x] compare and check off the data fields (below) against the ERD. Note any fields in the ERD that's missing from the list. (see Resources)
- [x] create a single model in Django (defining schema)
- [x] compare the check off associated tables (below) against the ERD. Note any associated tables in the ERD that's missing from the list (see Resources)
- [ ] write a test for the relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project).
- [x] write an API end point
- [x] write API unit tests
- [x] document the endpoint
- [x] Release #64 from ice box

### Resources/Instructions

- See [People Depot Resources wiki page](https://github.com/hackforla/peopledepot/wiki/Resources-and-Links) for links
  - ERD
  - Table and Field Definitions
  - API Endpoint
- See the [wiki page](https://github.com/hackforla/peopledepot/wiki/Create-Table-issues-data-gathering-workflow) for instructions.

### Description

Dictionary of languages

### Data Fields

1. Copied from spreadsheet and checked off according to ERD. (unchecked items indicate a mismatch between ERD and spreadsheet, which requires a review)

    - [x] (PK) id - int - Record Id
    - [x] name - varchar
    - [x] description - textfield
    - [x] logo - url varchar
    - [x] updated - varchar

1. In ERD only (having items here indicates a mismatch, which requires a review)
   - [ ] logo - URLField

### Associated Tables

1. Copied from spreadsheet and checked off according to ERD. (unchecked items indicate a mismatch between ERD and spreadsheet, which requires a review)
      - [x] project_language_xref (one-to-many)

1. In ERD only (having items here indicates a mismatch, which requires a review)
   - None

scarter31191:
@scarter31191@gmail.com

chelseybeck:
There is a way to do this that pulls in a list w/ IANA registry. See [django-language-field](https://pypi.org/project/django-language-field/)

fyliu:
Updated description field to be textfield in django.

AzaniaBG:
ETA: not sure, but I'm shooting for next Thursday, 7/13 (will update if needed).
Availability: 
[...]

PR #186, opened 2023-08-17
Create table language 54

AzaniaBG:


AzaniaBG:
Fixes #54 
* The how-to Guide indicates the read-only field is `updated_at` so I added that field instead of using the ERD's `updated_on` field.

PR #193, opened 2023-08-25
encountered this issue setting up local env so adding note for future…

chelseybeck:
I ran into this error setting up my local environment:
ERROR [web internal] load metadata for docker.io/library/python:3.10-bullseye

I am adding a note for future users.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -146,6 +146,18 @@ upstream        https://github.com/hackforla/peopledepot.git (push)
    ./scripts/buildrun.sh
    ```
 
+   Note: if you encounter an error regarding internal load metadata, you can run 
+   ```
+   sudo vi ~/.docker/config.json
+   ```
+   you should see a configuration such as
+   ``` 
+   {
+      "credsStore": "desktop.exe"
+   }
+   ```
+   update credsStore to credStore, save, and run the buildrun.sh script again.
+
 1. Create a super user for logging into the web admin interface
 
    ```bash


fyliu (approved):
Thanks for adding troubleshooting notes to the docs!

It looks like the naming of this key is related to how docker is installed[1], with the current official docs using "credsStore"[2].

Are you sure your renaming the key didn't just make it unrecognized and cause docker to revert to storing the plaintext passwords in the config.json file? You would see an "auths" section with the passwords encoded as "auth" keys if that were the case[3].

1. [post mentioning both keys](https://blog.adamgamboa.dev/docker-fixing-error-storing-credentials/)
2. [official docs on credentials store](https://docs.docker.com/engine/reference/commandline/login/#credentials-store)
3. [another guide explaining credential store](https://luiscachog.io/docker-login-the-right-way/)

fyliu:
After more thinking, I believe this is probably not the fix, so I'm going to close this PR.

Why?

1. It looks like Chelsey's config is changed back to the old `credsStore` key and I assume her docker is still working.

    I asked @chelseybeck what's in her config.json and it's this:
    ```json
    {
    	"auths": {},
    	"credsStore": "desktop",
    	"currentContext": "default"
    }
    ```
    Where the auths list is empty. It would not be a good idea to post the contents if there were values in there, since it contains essentially plain-text passwords to the docker account.

1. This error is what happens when there's a connection interruption to the docker image registry when it's trying to download the image. I suspect this might be the actual cause, but it would be temporary and any action could be interpreted as having fixed the problem.
    ```text
    ERROR [web internal] load metadata for docker.io
    ```
1. Changing the key from `credsStore` to `credStore` can be dangerous because it can disable the configuration and cause docker to store plaintext passwords in `auths` rather than in a secure credentials store. But what's likely to happen is docker detects there's no credsStore value set and prompts the user to set it. This is likely why @chelseybeck's config.json key got rewritten to the correct one after she changed it. Anyway, the plaintext behavior is demonstrated in the 3rd link I posted in a comment above.

PR #195, opened 2023-08-31
Create table language 54

AzaniaBG:
@fyliu ~ this is a backup I created after we squashed commits and rebased HfLA's `main` commits on top of the language feature branch. I think this should work. If this is acceptable, I will add the `precommit` checks.


fyliu (approved):
Thanks. This is rebased onto the main branch and works well.

One more commit cleanup suggestion: it would be great if you could move the script fix as the first commit and squash the other commits into the "...language model..." one.

AzaniaBG:
Thanks, @fyliu. I may not understand your goals for "clean" commits and would  like to discuss to improve my commits moving forward. Also, is this something you want to try to address in the Contributing guide?

fyliu:
Oh the branch got deleted.

I think we left this too long and didn't have the discussion that was referred to above.

I have a copy of that branch so I'll do the git manipulations and merge it into main. The overall code changes will be exactly the same but with different commits. They'll all have you as the author.

fyliu:
I manually pushed the commits to main. Here are the changes I made:

1. moved the script fix commit to the beginning, but separated the 0014 migration file from that commit since it belonged with the language feature.
1. squashed all the other commits into the feature commit and added the 0014 migration file
1. added new short messages to the commits and moved the original messages into the descriptions

PR #200, opened 2023-09-07
Removed requirements for python version in pre-commit

anandramakris:
Tested running pre-commit for black without specifying the python version. To do this, I created a different directory so black could be re-installed.

Was successful on my Windows 10 Home version 22H2.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,8 +2,6 @@
 # See https://pre-commit.com/hooks.html for more hooks
 default_install_hook_types: [pre-commit, pre-push]
 default_stages: [commit, push]
-default_language_version:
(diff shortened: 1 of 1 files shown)


fyliu (approved):
Thanks for contributing this!
It works for me using python 3.8, 3.10, and 3.11 on win11 wsl2 ubuntu. Python installed via pyenv. pre-commit installed in python 3.8 via pipx.
The only thing to note is it didn't show me an error even before the change. But if it solves it for you it'll help others.

If the pre-commit creator says it's not necessary, that's good enough for me. https://stackoverflow.com/questions/69462277/use-pre-commit-hook-for-black-with-multiple-language-versions-for-python

Issue #24, opened 2022-08-21
Create Table: permission_type


Neecolaa:
### Dependency
- #147 

### Overview

We need to create the permission_type table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project. 

#### Details

A table and a model are the same thing

### Action Items

- [x] identify and document table description (see spreadsheet under Resources)
  - [x] #176
  - Long Description: i.e., adminVrms, adminBrigade, adminProject, member
  - Sheet: PermissionType - Data

id | name | description
-- | -- | --
  | adminGlobal | Can CRUD anything
  | adminVrms | Can R anything and update permissions for users to the level of Admin Brigade
  | adminBrigade | Can CRUD anything related to a user or project assigned to their brigade
  | adminProject | Can Read and Update anything related to their assigned project
  | memberProject | Can Read anything related to their project that is visible
  | memberGeneral | Can express interest in a project's open role if they are the role

- [x] compare and check off the data fields (below) against the ERD. Note any fields in the ERD that's missing from the list. (see Resources)
- [x] compare the check off associated tables (below) against the ERD. Note any associated tables in the ERD that's missing from the list (see Resources)
  - [ ] `permission` table is a related table and is not yet created
	- [ ] #22 seems to require some further investigation
	- [ ] as mentioned in #147, there seems to be a need to further investigate & get clarity on how to implement permissions
  - [ ] `permission_history` table is a related table and is not yet created
	- [ ] #37 will need to investigate further later
- [x] create a single model in Django (defining schema)
- [ ] add seed data
- [ ] write a test for the relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project).
- [x] write an API end point  
![image](https://github.com/hackforla/peopledepot/assets/86270711/91b131b2-1808-4682-bd06-ff98467b8955)

- [x] write API unit tests
  - [x] To resume at [Adding API Tests](https://github.com/hackforla/peopledepot/blob/main/docs/how-to/add-model-and-api-endpoints.md#add-api-tests)
- [ ] document the endpoint

### Resources/Instructions

- See [People Depot Resources wiki page](https://github.com/hackforla/peopledepot/wiki/Resources-and-Links) for links
  - ERD
  - Table and Field Definitions
  - API Endpoint
- See the [wiki page](https://github.com/hackforla/peopledepot/wiki/Create-Table-issues-data-gathering-workflow) for instructions.

### Description

Dictionary of permission types used in #22 

[...]

PR #201, opened 2023-09-17
24 permission types alaw

alexlaw528:
Addresses #24. Continuation of work done by @joey-ma


fyliu (changes requested):
Looks good!

A couple of things:

1. Looks like you added a .DS_Store file in the merge commit. Otherwise, the conflict resolution looks really good.
2. You need to set up pre-commit for it to do some of the code auto-formatting. The contributing docs has instructions but they're missing a note about pipx. It's better to [install pipx](https://github.com/pypa/pipx) and then use it to install pre-commit. It will install pre-commit in a way that doesn't pollute your system global python environment. It's like what npx does in the javascript world.

I think it would be good for you to do these yourself, but I would also fix these in what I need to do to the commits, so you don't really need to do them if you don't want to.

I'll need to rebase the commits to `main` in order to keep Joey's contributions. I'm interested in keeping them because it's his first code contribution.

fyliu (changes requested):
Thanks for making the changes!

I was originally thinking of pushing all the commits into `main`, which meant all the commits should be working code, because we don't want to have intentionally broken commits in `main`, but that seems a lot of effort and not make too much sense to do. Instead, can you squash Joey's commits from 5/12 and 5/13 into the first commit and make sure the result of that passes pre-commit and works? Then add the changes you made and make sure that works. Just amend the commits with any fixes rather than have separate fixer commits.

Maybe write a new summary line for Joey's commits and list his changes in the detail text of the commit. 

I'm not sure about your preferred tool, but in the command line, `git rebase -i HEAD~20` and changing all but the earliest 5/12 Joey line prefixes to `s` should do the squash and ask you for a new commit message for the combined commit.

I'm not sure if this is the best way to do this. I think you may need a 2nd terminal to do and add the fixes to git staging while the 1st one is stopped waiting for the commit message.

Here are the authors and timestamps for the commits now.
Screen Shot 2023-09-20 at 11 29 37 PM

fyliu (approved):
Looks good!

I only tested that it builds and runs from scratch and ran the pre-commit checks. The code is identical to the last time I reviewed it except for removing an extra file.

PR #213, opened 2023-10-05
Updated CONTRIBUTING.md file

### LINKED ISSUE #211: Update the Contributing Guide
We need to add a comment in section 2.1.4 of the contributing guide, instructing Windows users to install WSL and Git inside it to prevent or resolve errors.

Windows users can have errors when running Docker scripts, possibly caused by line-ending definitions or PATH conflicts. 

[CONTRIBUTING 2.1.4](https://github.com/hackforla/peopledepot/blob/main/docs/CONTRIBUTING.md#214-install-git)

benpinhassi:
Fixes #211 

I've updated the CONTRIBUTING.md section -  2.1.4 Install Git.

In continuation of our conversation on errors when running scripts while using Git Bash on Windows.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -34,7 +34,14 @@ Set up two-factor authentication on your account by following this [guide](https
 
 Before cloning your forked repository to your local machine, you must have Git installed. You can find instructions for installing Git for your operating system [**here**](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
 
-For Windows, you may want to [install a Windows Subsystem for Linux (WSL)](https://code.visualstudio.com/docs/remote/wsl) and then install Git from WSL.
+Installation Guide for Windows Users
+- we recommend [installing Windows Subsystem for Linux (WSL)](https://code.visualstudio.com/docs/remote/wsl). WSL provides a Linux-compatible environment that can prevent common errors during script execution.
+- After setting up WSL, install Git directly from the Linux terminal. This method can help avoid complications that sometimes arise when using Git Bash on Windows.
+- If you prefer Git Bash or encounter errors related to line endings when running scripts, the problem might be due to file conversions in Windows. To address this, configure Git as follows: 
+```bash
+    git config --system set autocrlf=false
+```
+<strong><em>Feel free to reach out in the [Hack for LA Slack channel](https://hackforla.slack.com/messages/people-depot/) if you encounter any errors while running scripts on Windows. </em></strong>
 
 Please note that if you have a Mac the page offers several options (see other option, if you need to conserve hard drive space) including:
 


docs/CONTRIBUTING.md
```diff
@@ -34,7 +34,14 @@ Set up two-factor authentication on your account by following this [guide](https
 
 Before cloning your forked repository to your local machine, you must have Git installed. You can find instructions for installing Git for your operating system [**here**](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).
 
-For Windows, you may want to [install a Windows Subsystem for Linux (WSL)](https://code.visualstudio.com/docs/remote/wsl) and then install Git from WSL.
+Installation Guide for Windows Users
+- we recommend [installing Windows Subsystem for Linux (WSL)](https://code.visualstudio.com/docs/remote/wsl). WSL provides a Linux-compatible environment that can prevent common errors during script execution.
+- After setting up WSL, install Git directly from the Linux terminal. This method can help avoid complications that sometimes arise when using Git Bash on Windows.
+- If you prefer Git Bash or encounter errors related to line endings when running scripts, the problem might be due to file conversions in Windows. To address this, configure Git as follows:
```
AzaniaBG, line 40: This sentence on line 40 is a bit ambiguous and I wonder if we should provide the error Ben posted in Slack for context. I believe the git config solution is for an error that occurs when using Git Bash to install Linux (instead of WSL). I'm assuming it won't hurt to configure Git as instructed here, so this PR is probably fine as-is. @fyliu, thoughts on this?

AzaniaBG (commented):
Ben is unable to make changes (see his Slack comment on 10/8/23), so this PR should be sufficient, however, I wonder if it would be better to specify the type of error you would see that would warrant using the provided git config solution - see my inline comment on line 40 for details.

AzaniaBG:
I moved this status to "PR being reviewed" and added @fyliu as a reviewer (in addition to myself).

docs/CONTRIBUTING.md
fyliu: @AzaniaBG you're right for both things. It's okay to run the command in wsl. The git config command is for git bash users. I guess it's okay as-is since the user is probably using git bash already if they encounter this. This bullet is for people that really want to use git bash despite the recommendation.

After the config change, they might need to re-clone or re-check out the branch. I'm not sure what's needed. I can try it on my win 11 machine after this and report back.

fyliu: I tried git reset and it didn't seem to work. I then tried git switch main to switch to the same branch, but it didn't work. Both commands went through and modified all the files.

fyliu: After a discussion at the last meeting. I'm okay with it not having a lot of details on how to fix windows git bash, since it's the non-preferred way.

fyliu (approved):
Looks good!

PR #222, opened 2023-10-30
Create prod Dockerimage from alpine base image

### LINKED ISSUE #177: Switch to alpine base image in Dockerfile
### Overview

The current PD docker image is 1.3GB. We should switch to alpine so that we can reduce the image size.

### Action Items

- [x] Use the latest alpine base image
- [x] Convert rest of Dockerfile to alpine, such as using apk instead of apt

### Post Done

- [ ] take #302 out of ice box and put it in New Issue Review

### Resources/Instructions

- See [docker-mkdocs code](https://github.com/hackforla/docker-mkdocs) for ideas.

alexlaw528:
Fixes #177 

- Added a production Docker image using alpine as base-image.

- Current Docker image is meant for development. It utilizes a larger base-image and contains dependencies unused in production.
- Reduced image size from 1.26 GB to 392.06 MB
Screenshot 2023-10-29 at 7 29 38 PM

### Deploy prod environment locally
- Run following commands from` /peopledepot` root directory:

- Build production image as "alpine-pd"
`docker build -t alpine-pd -f ./app/prod/Dockerfile ./app/prod`

- Start up production services 
`docker-compose -f docker-compose.prod.yml up`

### DIFF
--- a/app/prod/Dockerfile
+++ b/app/prod/Dockerfile
@@ -1,0 +1,21 @@
+FROM python:3.11-alpine
+
+# set work directory
+WORKDIR /usr/src/app
+
+# set environment variables
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+# install system dependencies
+RUN apk update && \
+    apk add --no-cache netcat-openbsd gcc postgresql graphviz musl-dev libffi-dev
+
+# install dependencies
+RUN pip install --no-cache-dir pip~=21.3.1
+COPY ./requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+# # copy entrypoint.sh
+ENTRYPOINT ["./entrypoint.sh"]
+
--- a/app/prod/requirements.txt
+++ b/app/prod/requirements.txt
@@ -1,0 +1,14 @@
+black~=22.3.0
+cryptography~=37.0.2
+Django~=4.0.1
+django-extensions~=3.1.5
+django-linear-migrations~=2.12.0
+django-phonenumber-field[phonenumbers]~=6.3.0
+django-timezone-field==5.0
+djangorestframework~=3.13.1
+drf-jwt~=1.19.2
+drf-spectacular==0.22.1
+psycopg2-binary~=2.9.3
+pydot~=1.4.2
+pytz==2022.1
+tzdata==2022.1
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,0 +1,27 @@
+version: '3.8'
+
+services:
+  db:
+    image: postgres:13.0-alpine
+    volumes:
+      - postgres_data:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_USER=people_depot
+      - POSTGRES_PASSWORD=people_depot
+      - POSTGRES_DB=people_depot_dev
+    container_name: people_depot_db
+  web:
+    build: ./app/prod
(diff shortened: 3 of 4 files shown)


fyliu:
@alexlaw528 I think the "alpine" is usually part of the image tags. Like `-t pd:latest-alpine`

I'm not sure if there's a best practice for managing dev and prod dockerfiles. I see some people do Dockerfile.prod, or use the same Dockerfile for both with conditionals. I guess putting it in a directory is as good as any. We can think about this later.

We should also do multistage build to maybe avoid part of the 211MB that system dependencies take up. This is just a note for a later issue.

app/prod/requirements.txt
```diff
@@ -0,0 +1,14 @@
+black~=22.3.0
```
fyliu, line 1: I would remove black, drf-spectacular, pydot, and maybe django-linear-migrations from the production build. I don't think there's any use for them outside development.

docker-compose.prod.yml
```diff
@@ -0,0 +1,27 @@
...
+    container_name: people_depot_db
+  web:
+    build: ./app/prod
+    command: python manage.py runserver 0.0.0.0:8000
+    volumes:
+      - ./app/:/usr/src/app/
+    ports:
+      - 8000:8000
+    env_file:
+      - ./.env.dev
```
fyliu, line 21: Maybe have a separate .env or .env.prod file for production

docker-compose.prod.yml
```diff
@@ -0,0 +1,27 @@
...
+services:
+  db:
+    image: postgres:13.0-alpine
+    volumes:
+      - postgres_data:/var/lib/postgresql/data/
+    environment:
+      - POSTGRES_USER=people_depot
+      - POSTGRES_PASSWORD=people_depot
+      - POSTGRES_DB=people_depot_dev
+    container_name: people_depot_db
```
fyliu, line 12: Setting the container name might not be a good idea. I'll have to look up the reasoning for this again.

fyliu: Oh yeah, the container_name allows you to refer to it by name in a `docker` command. If you intend to refer to it using docker-compose, like `docker-compose run db...` then you only need to use the service name. Setting the container name also restricts docker-compose from generating multiple instances of it. I'm not sure if there's an advantage to that or not. So the reasoning for not setting it is really that it's not very useful.

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - 8000:8000
     env_file:
       - ./.env.dev
+    container_name: people_depot_dev
```
fyliu, line 13: Same reasoning as above for not setting the container name. We're mainly referring to containers using `docker-compose`.

app/prod/Dockerfile
fyliu, line 1: This is probably fine as a conversion from apt to apk. It's missing some production things that we can add in other issues. Multi-stage build, --no-compile flag, maybe apk cache cleanup.

fyliu (changes requested):
Thanks for working on this. Please excuse some of the comments that read more like I'm not sure if it needs a change. I think there's some miscommunication in the issue. The issue was intended to convert the development Dockerfile to use alpine and apk. It's okay to do it for production too. I see you made copies of some files and removed the development dependencies. It's just that it's missing elements that I would consider to be needed for production, so I'm not sure if I should ask for them since they're not in the issue.

Or would it be okay if you just make the change to the development setup and then do the same for production? I'm not sure what the best way to do this is.

ExperimentsInHonesty:
- We are closing this and it's being replaced by #275

PR #228, opened 2023-12-05
Changes for kb

ethanstrominger:
Fixes #replace_this_text_with_the_issue_number

-
-
-

-
-
-

<details>

image

</details>

<details>

image

</details>


app/core/api/serializers.py
```diff
@@ -202,8 +202,8 @@ class ProgramAreaSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ProgramArea
```
ethanstrominger, line 204: Go back to uuid

app/core/api/urls.py
```diff
@@ -34,7 +38,11 @@
 )
 
 urlpatterns = [
-    path("me/", UserProfileAPIView.as_view(), name="my_profile"),
+    path('secure-api/getusers', SecureApiView.as_view(), name='secure_api_get_users'),
+    path('secure-api/newcognitouser', SecureCreateUserFromCognito.as_view(),name='secure_new_user'),
+    path('me/', UserProfileAPIView.as_view(), name='my_profile'),
+    path('home/', TemplateView.as_view(template_name='common/home.html'), name='home'),
```
ethanstrominger, line 44: fix this one

fyliu (changes requested):
Maybe it's too early for me to review this since several things aren't complete yet.

Looks like there's a lot of directions to the changes. Some of them might be better split into their own PRs to review separately.

1. cognito auth client in the backend
    - missing instructions on how to use it. What url to use. Needing to create the client id in AWS and put it in the .env file.
1. things not needed
    - changing model id between int and uuid (it's also causing existing tests to fail)
    - changing the .env file name
    - envsettings.sh file
1. shouldn't be needed
    - API for peopledepot clients to create new peopledepot users
    - default assigning staff to new users allows all users to use the admin site
1. scripts and instructions for running the backend without docker

app/templates/account/login.html
fyliu, line 1: I think it's `accounts/login.html`. Accounts with an s at the end.

PR #243, opened 2024-01-18
Local env

### LINKED ISSUE #240: local deployment without docker
### Overview

We would like to support local deployment of the backend without the use of docker. This would support a greater number of developers who are more comfortable running things in virtual environments.

### Action Items

- [ ] set up the project to run by itself
- [ ] document how to run it

ethanstrominger:
- Fixes #240

- moved the .env file from [root]/.env.dev to [root]/app/.env.docker
- scripts and instructions on running the backend on a virtual env

- It's necessary to move the .env file into the app/ directory for the virtual env
- We want to have a way for devs to run the backend without docker if they needed to

<details>

image

</details>

<details>

image

</details>


ethanstrominger:
pre-commit.ci autofix

fyliu (changes requested):
Env file

With the .env.docker-example file and .env.local-example files, it's easy for them to be out of sync. Right now, I see one of them has the `us-east-2` user pool and the other with `us-west-2` user pool. The project itself should probably use the `us-west-2` one. This might be where the wiki can provide alternate configurations. 

It would be nicer if the `.env.docker-example` and `.env.local-example` could be combined into a single file in `app/`, and the `docker-compose.yml` file changed to point to it. Maybe even just name it `.env-example` or back to `.env.dev-example`.


---
Docs

Here are the steps that led to a good outcome on my machine. Maybe the docs can include some of it.
```
cd app

cp .env.local-example .env.local
python -m venv venv

source ../scripts/activate.sh
pip install -r requirements.txt

../scripts/start-local.sh

deactivate
```

---

- I'm not able to see a good reason for adding container names in the docker-compose file here.
- the `docs/CONTRIBUTING.md` file is still useful
- you should probably do what some of the shellcheck rule want. The check status one seems doable.

fyliu:
So [here](https://github.com/fyliu/peopledepot/commit/eb8cf5b1d0de957e0660909b5e4c30d8673ab47f) I modified the code to satisfy the shellcheck warning.

fyliu:
@ethanstrominger I've been trying to clean up the commit, but it looks like something is causing the local setup to use sqlite even though only the postgres settings are present.

fyliu:
I'm not sure if I'm missing a step. Here's what I did
```
peopledepot/app
❯ cp .env.local-example .env.local                                                                                                                                                     
❯ python -m venv venv
❯ source ../scripts/activate.sh
Sourced OK                                                                                                                                                                                                       
Done                                                                                                                                                                                                                                                                                                                                                                                                                              
❯ pip install -r requirements.txt
...
❯ source ../scripts/start-local.sh
SQL USER people_depot
file = SQL_ENGINE= / SQL_ENGINE= /
Loading environment variables from SQL_ENGINE=
File SQL_ENGINE= not found
If executing locally, copy .env.example.dev to SQL_ENGINE= and edit as needed
ERROR: loadenv.sh failed
```

fyliu:
Added some change descriptions to this PR.

fyliu (approved):
I think this is good now. The only breaking change for devs is they have to move their `.env.dev` to `app/.env.docker` in order for their project to continue working.

PR #270, opened 2024-03-18
Create table project_sponsor_partner_xref

### LINKED ISSUE #65: Create Table: project_sponsor_partner_xref
### Overview

We need to create the project_sponsor_partner_xref table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items

  - [x] if not, reach out to PD leads
  - [x] add the docstring from the section below
- [x] write an API end point
- [x] write API unit tests
- [ ] document the endpoint

### After Merge

- [ ] Release #182 from the ice box if this is the last dependency

### Resources/Instructions

  - ERD
  - Table and Field Definitions
  - API Endpoint

### Sponsor/Partner clarification

Sponsor/partner relationships stored in this table are project-dependent. They cannot be both a sponsor and a partner for the same project, so if is_sponsor is FALSE they are a project partner, if is_sponsor is TRUE they are a project sponsor.

### Description

Stores the relationship between sponsors/partners and projects (Joins tables sponsor_partner and project)

### Table name vs Model name

- [ ] The table name should be `project_affiliate_xref`, but the Django model name should be `Affiliation` for ease of coding.
  - [ ] Use `db_table` field in meta to specify the db table name

### Data Fields

    - [x] (FK) project_id - int - id of project linked to partner
    - [x] (FK) partner_id - int - id of partner linked to project
    - [x] created - timestamp
    - [x] ended - timestamp
    - [x] is_sponsor - boolean
    - [x] is_partner - boolean

   - None

### Associated Tables

   - [x] project (many-to-one)
   - [x] sponsor_partner (many-to-one)

   - None

freaky4wrld:
Fixes #65 , #272

- Created a model for the `ProjectSponsorPartnerXref`
- Registered the model to adminSite
- Created API endpoints
- Renamed the model `sponsor_partner` to `affiliate`

- To add a table `ProjectSponsorPartnerXref` in the database
- To make the model name shorter to `affiliate`


fyliu (changes requested):
Look great! Just a few things I noticed in the code should be changed.

app/core/models.py
```diff
@@ -329,3 +329,23 @@ class Sdg(AbstractBaseModel):
...
+    Sponsor/partner relationships stored in this table are project-dependent.
+    They cannot be both a sponsor and a partner for the same project,
+    so if is_sponsor is FALSE they are a project partner,
+    if is_sponsor is TRUE they are a project sponsor.
+    """
+
+    partner_id = models.ForeignKey(SponsorPartner, on_delete=models.CASCADE)
+    project_id = models.ForeignKey(Project, on_delete=models.CASCADE)
+    ended_at = models.DateTimeField("Ended at", null=True, blank=True)
+    is_sponsor = models.BooleanField(null=True)
```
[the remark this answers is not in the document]
freaky4wrld: Yep totally agreed, tell me when the issues are ready!!

freaky4wrld:
@fyliu the changes you suggested are done, except the one with the `is_sponsor` one.

app/core/models.py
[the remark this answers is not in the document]
fyliu: @freaky4wrld Can you just make the field required for now.

There's another thing I didn't consider since I'm not really a database person. Do we need the partner-project association to be unique? I'm thinking yes and that we need to add a uniqueness constraint to the model.

fyliu (changes requested):
Thanks for making the changes. I left another comment on the model and thought of something new. Can you take a look.

freaky4wrld:
@fyliu the changes requested are done please check

app/core/models.py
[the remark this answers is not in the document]
fyliu: Thanks! I think we do need to add a uniqueness constraint. Here's the generated suggestion as I was trying to add it myself. It seems to work when I tried to add the same relationship twice. I looked up the `unique_together` constraint that I remember hearing about, but the django docs now recommends using this more generic syntax.
```python
class ProjectSponsorPartnerXref(AbstractBaseModel):
    ...

    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=["project", "partner"], name="unique_project_partner"
            )
        ]
```

fyliu:
I'm realizing that this PR is essentially ready but we were discussing changes to it over slack chat. I'm pasting that here for reference

- updating the table name to `project_affiliate_xref`
- updating the model name to `Affiliation`, specifying that the `db_table` is the name above
- converting the boolean to an enum choice of `sponsor` or `partner`
- maybe also changing the `sponsor_partner` table name to `affiliate` and the FK field in `Affiliation`

freaky4wrld:
@fyliu I've made the suggested change, have a look and do resolve the issue I've mentioned in slack!!

app/core/models.py
```diff
@@ -182,7 +182,10 @@ class SponsorPartner(AbstractBaseModel):
     is_org_sponsor = models.BooleanField(null=True)
     is_org_partner = models.BooleanField(null=True)
 
-    # PK of this model is the ForeignKey for project_partner_xref
+    # PK of this model is the ForeignKey for project_affiliate_xref
+
+    class Meta:
+        db_table = "affiliate"
```
fyliu, line 188: I actually meant to change both the model name and the table name to `affiliate`.

app/core/models.py
```diff
@@ -329,3 +332,39 @@ class Sdg(AbstractBaseModel):
...
+
+    AFFILIATION_CHOICES = [
+        (SPONSOR, "Sponsor"),
+        (PARTNER, "Partner"),
+    ]
+
+    affiliate = models.ForeignKey(SponsorPartner, on_delete=models.CASCADE)
+    project = models.ForeignKey(Project, on_delete=models.CASCADE)
+    ended_at = models.DateTimeField("Ended at", null=True, blank=True)
+    is_sponsor = models.CharField(max_length=20, choices=AFFILIATION_CHOICES)
```
fyliu, line 356: I think this is the type of affiliation. But `type` is a python builtin word so we have to choose something else like `kind` or another word.

fyliu: I was originally thinking of using python enums as choices. There's some advantages to using enums.

fyliu (changes requested):
Thanks for the PR! The migrations applied fine for me.

There a few extra commits here that I think you brought over from `main`. They're making the github's PR changes view look more complicated. We should figure out how to make it cleaner for future PRs.

You made good assumptions with little information, but some of the things I just need to specify better. I think I need to create those other issues to better communicate what's needed.

app/core/admin.py
```diff
@@ -138,8 +139,8 @@ class PracticeAreaAdmin(admin.ModelAdmin):
     )
 
 
-@admin.register(SponsorPartner)
-class SponsorPartnerAdmin(admin.ModelAdmin):
+@admin.register(Affiliate)
+class SponsAffiliate(admin.ModelAdmin):
```
fyliu, line 143: This should be called `AffiliateAdmin`

app/core/models.py
```diff
@@ -329,3 +333,42 @@ class Sdg(AbstractBaseModel):
...
+            return self.value
+
+        @classmethod
+        def choices(cls):
+            return [(key.value, key.name) for key in cls]
+
+    affiliate = models.ForeignKey(Affiliate, on_delete=models.CASCADE)
+    project = models.ForeignKey(Project, on_delete=models.CASCADE)
+    ended_at = models.DateTimeField("Ended at", null=True, blank=True)
+    is_sponsor = models.CharField(max_length=20, choices=AffiliationChoices.choices())
```
fyliu, line 360: This should be called something like `affiliation_type` so that code can do
```
if self.affiliation_type == "sponsor"
```
in line 371

fyliu (changes requested):
Thanks for making the changes! I just have a couple small things.

fyliu:
@freaky4wrld I just realized that the boolean to enum change hasn't been agreed on with @Neecolaa yet. Will have to bring that up for discussion in the next meeting.

freaky4wrld:
> @freaky4wrld I just realized that the boolean to enum change hasn't been agreed on with @Neecolaa yet. Will have to bring that up for discussion in the next meeting.


@fyliu did you had the discussion with @Neecolaa. I'm still waiting for the confirmation and further info on changes to be made!!

Neecolaa:
We're adding a row:
`is_partner` is a boolean that indicated whether or not the given affiliate is a partner for the given project. Projects can now have an affiliate be both a partner and sponsor. We now have definitions for the affiliate types as well:
**A Partner** is an entity that's doing ongoing work with us on the project.
**A sponsor** is someone who has provided funds or an in-kind donation.

We won't be using an enum, just the two booleans.

Neecolaa:
@freaky4wrld see update comment above

freaky4wrld:
@fyliu and @Neecolaa the changes have been made as suggested, please check and provide feedback
Thanks

app/core/api/views.py
```diff
@@ -314,3 +316,17 @@ class SdgViewSet(viewsets.ModelViewSet):
     permission_classes = [IsAuthenticated]
     queryset = Sdg.objects.all()
     serializer_class = SdgSerializer
+
+
+@extend_schema_view(
+    list=extend_schema(description="Return a list of all the recurring events"),
```
fyliu, line 322: These should have the "recurring events" replaced with "affiliations"

app/core/models.py
```diff
@@ -182,7 +182,10 @@ class SponsorPartner(AbstractBaseModel):
     is_org_sponsor = models.BooleanField(null=True)
     is_org_partner = models.BooleanField(null=True)
 
-    # PK of this model is the ForeignKey for project_partner_xref
+    # PK of this model is the ForeignKey for project_affiliate_xref
+
+    class Meta:
```
fyliu, line 187: No need to specify `db_table` if the name is the same as the model name

app/core/api/serializers.py
```diff
@@ -296,3 +297,21 @@ class Meta:
...
+    Used to retrieve Affiliation
+    """
+
+    class Meta:
+        model = Affiliation
+        fields = (
+            "uuid",
+            "affiliate",
+            "project",
+            "ended_at",
```
fyliu, line 313: I think we need to show the "created_at" for this to make sense. Right now, it only shows "ended_at" which looks weird without the begin datetime.
image

fyliu (changes requested):
Thanks for making the update and sorry for the change in requirement.

I found a couple things that could be improved. I'm short on time today so I'll continue tomorrow.

app/core/api/views.py
```diff
@@ -314,3 +316,17 @@ class SdgViewSet(viewsets.ModelViewSet):
     permission_classes = [IsAuthenticated]
     queryset = Sdg.objects.all()
     serializer_class = SdgSerializer
+
+
+@extend_schema_view(
+    list=extend_schema(description="Return a list of all the affiliations"),
+    create=extend_schema(description="Create a new recurring event"),
```
fyliu, line 323: These strings need to be updated also. Sorry I was lazy about selecting all the lines.

fyliu (changes requested):
Thanks for the updates.

I think I wasn't clear enough about the `created_at` field. It's already provided by the `AbstractBaseModel` and you just have to include it in the `AffiliationSerializer`. Can you remove it from the Affiliation model class. That should make it use the inherited one. Thank you!

fyliu (approved):
Works great, thanks!

I tried to simplify the commit history but it looks like I have to just squash them down to a single commit because of 2 models being changed. I also added a few other fixes that you can see at the end of the branch history in my [forked branch](https://github.com/fyliu/peopledepot/tree/create-table-project-sponsor-partner-xref-65)

PR #273, opened 2024-04-01
Eliminate external JSON export script

### LINKED ISSUE #144: Eliminate external JSON export script
### Dependency
- PR #141, which introduces the external script

### Overview
We're using an external Goggle Apps script to export initial data from the spreadsheet in JSON form, but the functionality can be brought in to our own script and eliminate the manual step.
 - from #36, which uses the external script

### Action Items

- [x] Create a Python script to read sheets data
  - [ ] Read all the data from a sheet by sheet name
- [x] Combine or call the script from the initial data conversion script
  - [x] Make sure the new setup works for SOC Major and User Status data
  - [ ] Bonus: call the conversion script with the model name as input and have it output to a file named after the model
- [x] Remove external script and associated docs

### Resources/Instructions
- [Google Sheets python quickstart](https://developers.google.com/sheets/api/quickstart/python)

anandramakris:
Fixes #144 

- Changed convert script so it gets data from a csv spreadsheet directly. The data is converted to JSON inside the script, removing the step of having to run the Google Apps script for JSON data.
- Made sure id and uuid values are cast to int rather than string while converting from JSON to Python.
- Modified documentation to show how to export the spreadsheet to csv.

- Reading directly from the downloaded spreadsheet is simpler than copying the JSON data from the Google Apps script into a new file.

<details>

image

</details>

<details>

image

</details>

### DIFF
--- a/.github/ISSUE_TEMPLATE/add-seed-data-to-a-table.md
+++ b/.github/ISSUE_TEMPLATE/add-seed-data-to-a-table.md
@@ -2,9 +2,9 @@
 name: Add seed data to a table
 about: For adding data that will be the same no matter what organization uses People
   Depot
-title: Add seed data for [Replace with TABLE NAME]
-labels: 'Feature: Table Creation, good first issue, missing: feature, missing: stakeholder,
-  role: back end, size: 1pt'
+title: 'Add seed data for Table: [Replace with TABLE NAME]'
+labels: 'Feature: Table Creation, good first issue, missing: feature, role: back end,
+  s: CTJ, s: hackforla.org, s: kb, s: org, s: PD team, s: tables, s: VRMS, size: 1pt'
 assignees: ''
 
 ---


fyliu (changes requested):
Great work. It's much easier to convert the data in to scripts. Works great for the example model in the documentation and the documentation is easy to follow.

There's some problem in the convert script that cases fields to mismatch when some of the fields are empty. You can see it on "WinType - Data"
- I'm not too concerned about the extra text comments. We can just say that the prerequisite is that there can only be data in these sheets, and comments should be done as spreadsheet comments or something.
- The blank fields do need to become blank strings in the script.

app/scripts/convert.py
```diff
@@ -47,7 +48,23 @@ def to_keys_indexes_str(_input):
...
+        headers = []
+        headers_row = 1
+        for row in csv_data:
+            if headers_row == 1:
+                headers = row
+                headers_row = 0
+            else:
+                row_data = {}
+                for i in range(0, len(row)):
+                    if row[i]:
```
fyliu, line 63: This looks to be the condition that incorrectly skips empty fields. Is this solving another problem?

fyliu (changes requested):
I think I found the code that's causing the problem.

app/scripts/convert.py
```diff
@@ -47,7 +48,23 @@ def to_keys_indexes_str(_input):
 
 def convert(file_path):
```
fyliu, line 49: You can tell flake8 to ignore this line with inline comment `# noqa: C901`
The `C901` is from the pre-commit check near the bottom of the PR.
[Reference](https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html)

fyliu (changes requested):

app/scripts/convert.py
fyliu: I see what it's doing. If `id` is blank, it doesn't include it and allows django to generate it.

Maybe the right way to resolve this is to say something like all fields must be populated.

PR #276, opened 2024-04-19
Update technology table #204

### LINKED ISSUE #204: Update Table: technology (rename to stack_element)
### Dependency
 - #207 
 - #206 

### Overview
We need to update the technology table model to become the new stack_element table. We are consolidating language, technology, and tool into one table.

#### Details
- The initial model issue is #39

- Discussion leading to this change is here: #202

### Action Items
- [x] Update existing Django model
- [ ] Write a test for the new relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project) if any.
- [ ] Update API end point
- [ ] Update API unit tests
- [ ] Document the endpoint in Swagger
- [ ] release #205 from ice box

#### Columns to Remove

- None

#### Columns to Add

- [x] (FK) `stack_element_type_id` - int (see the Event model in models.py for how to do this. Also see discussion below on naming this to `type`)

#### Additional Changes

- [x]  Change model name from technology to stack_element

### Resources

  - ERD
  - Table and Field Definitions
  - API Endpoint

#### Model relationships

All the code here are not guaranteed to be correct or have correct syntax. It's just an explanation with example "code".

It's probably going to look something like this
`type = models.ForeignKey(StackElementType, on_delete=models.CASCADE)`

The `type` name is so that when there's a `StackElement` object like `language_python = StackElement.objects.create("python")`, we can do `python_language.type = type_language`, where `type_language` is an object of StackElementType with the name "language".

The point is that `python_language.type` is easier to use than `python_language.stack_element_type_id`. In Django, the convention is to drop the `_id` for relationships.

#### Testing ideas

The `stack_element_type` object would hold something like "language", "technology", or "tool". The `stack_element` object holds the name of something in one of those categories.

##### Create a python language stack element

- create a `stack_element_type` object of "language"
- create a `stack_element` object of "python" with the "language" type
- check the `stack_element`'s __str__ method
- check that the `stack_element`'s type is "language"
```mermaid
classDiagram
direction LR
python --> language : has type
```

AzaniaBG:
Fixes #204 

- Renamed Technology model to StackElement
- Added element_type FK field to StackElement, pointing to a StackElementType
- Updated graphviz version

- The requirements are part of #204
- graphviz had to be updated for the Docker image to build


fyliu:
@AzaniaBG I finally figured out the problem. What you have is nearly done except you missed the step of adding `element_type` as a field to the serializer. It's saying it's null because the model is not receiving it from the API call, because the serializer doesn't know about it.

If you have a little bit of time, it might be easier to explain it at a meeting if what I wrote is confusing.

fyliu:
Setting this PR for review. It's based on old code, so it needs updating to the current `main` before applying the fix. I think the best way is to squash the original PR and rebase it on the current `main`, then resolve conflicts.

fyliu:
The [combined, rebased commit, with fixes and migration file update branch](https://github.com/fyliu/peopledepot/tree/pr/276-fixed) is in my fork.

I combined and force-pushed the updated commit to @AzaniaBG's branch to update this PR. It'd be good if someone else can try at least running it to make sure it runs.

dmartin4820:
Pull the changes from this PR to my local, ran scripts/buildrun.sh and didn't have any issues. 

I have no other comments on the commits either.

dmartin4820 (approved):

PR #303, opened 2024-06-09
Add seed data permission type 278

freaky4wrld:
Fixes #replace_this_text_with_the_issue_number

- added json file and migrations file for the seed data
- modified `Dockerfile` to install postgresql during docker-build

- For setting up the initial data for `PermissionType` model
- Postgresql is necessary for seeding the data

### DIFF
--- a/app/data/migrations/0004_permissiontype_seed.py
+++ b/app/data/migrations/0004_permissiontype_seed.py
@@ -1,0 +1,24 @@
+from django.db import migrations
+
+from core.models import PermissionType
+
+
+def run(__code__, __reverse_code__):
+   items = [
+        (1, "adminGlobal", "Granted to People Depo Admins. Can CRUD anything."),
+        (2, "adminVrms", "Granted to VRMS the application.  Can R anything and update permissions for users to the level of Admin Brigade"),
+        (3, "adminBrigade", "Granted to Brigage Leads.  Can CRUD anything related to a user or project assigned to their brigade "),
+        (4, "adminProject", "Granted to Product Managers.  Can Read and Update anything related to their assigned project"),
+        (5, "practiceLeadProject", "Granted when a user is promoted or assigned to be a Project Lead.  Can R and U anything related to people in their practice area (must attend lead events)"),
+        (6, "practiceLeadJrProject", "Granted when a user is promoted or assigned to be a Jr Project Lead.  Can R and U anything related to people in their practice area (should attend lead events)"),
+        (7, "memberProject", "Granted when a user joins a project.  Can Read anything related to their project that is visible"),
+        (8, "memberGeneral", "Granted when a user finishes onboaring.  Can express interest in a project's open role if they are the role"),
+    ]
+
+
+class Migration(migrations.Migration):
+    initial = True
+    dependencies = [("data", "0003_sdg_seed")]
+
+
+operations = [migrations.RunPython(run, migrations.RunPython.noop)]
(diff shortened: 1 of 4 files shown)


app/data/migrations/0004_permissiontype_seed.py
```diff
@@ -0,0 +1,52 @@
...
+            "memberProject",
+            "Granted when a user joins a project.  Can Read anything related to their project that is visible",
+        ),
+        (
+            8,
+            "memberGeneral",
+            "Granted when a user finishes onboaring.  Can express interest in a project's open role if they are the role",
+        ),
+    ]
+
```
fyliu, line 45: This is missing some code to insert the data.

Here's what the `sdg_seed` one has.
```python
    for uuid, name, description, image in items:
        Sdg.objects.create(uuid=uuid, name=name, description=description, image=image)
```

fyliu (changes requested):
Thanks for adding the seed data and finding the missing postgres dependency. It looks like it's missing some code that should have been pasted over. Can you take a look at my comment? Thanks!

freaky4wrld:
@fyliu added the missing lines in the file, just a question do we still need `postgresql` added, or we can skip that???

fyliu:
@freaky4wrld I think we should skip it. There's several other management commands dealing with databases but I think `dbshell` is the only one that needs `psql`.

fyliu:
@freaky4wrld I'm still having trouble seeing the data inserted into the database. Can you check if this happens to you too?

Here are my steps:
```bash
# build and run the container in the right branch
./script/buildrun.sh
# I also tried it with the -v option to delete and start from a fresh data volume

# make sure data/0004 migration is applied
docker-compose exec -T web python manage.py showmigrations data

# check permissiontype table data
./scripts/db.sh -c "select * from core_permissiontype;"
# shows empty table

# check sdg table has data
./scripts/db.sh -c "select * from core_sdg;"

# show core tables with extended info (size and desc)
./scripts/db.sh -c "\dt+ core*"
# shows that core_permissiontype is essentially empty

# more debug attempt: unapply and reapply the migration
./scripts/migrate.sh data 0003
./scripts/migrate.sh data 0004
# I checked and the table is still empty
```

freaky4wrld:
@fyliu yep the seed data doesn't display irrespective of the changes you made in the `db.sh`. Don't know how it happened but I made sure it shows the seed data in the `db-table`... should I run the script again to insert the data into the `db-table`

app/data/migrations/0004_permissiontype_seed.py
```diff
@@ -0,0 +1,54 @@
...
+    for uuid, name, description in items:
+        PermissionType.objects.create(uuid=uuid, name=name, description=description)
+
+
+class Migration(migrations.Migration):
+    initial = True
+    dependencies = [("data", "0003_sdg_seed")]
+
+
+operations = [migrations.RunPython(run, migrations.RunPython.noop)]
```
fyliu, line 54: I figured out why the migration file doesn't work. This line needs to be indented under the `class Migration`

fyliu (changes requested):
I found the problem and commented on the line. Please make the correction.

This is caused by a problem in the documentation to create initial data script. I will push the fix along with another fix to allow the migrations to be reversed.

fyliu:
@freaky4wrld can you make the necessary change to the code? It's just adding an indentation level for that line.

freaky4wrld:
@fyliu thanks for pointing out the error I've made the desired changes

fyliu (approved):
Looks great, thanks!

PR #314, opened 2024-06-17
fix(docs): create initial data scripts page

### LINKED ISSUE #313: docs: fix initial data doc
### Problem discovered in this issue

- #303

### Overview

We need to fix the `create initial data scripts` page.
The guide has a bug in the code which makes code created with the guide non-functional.
The data migration files aren't reversible, so we should fix that.

### Action Items

- [x] fix the guide

### Resources/Instructions

Here's [the page that contains the error](https://github.com/hackforla/peopledepot/blob/main/docs/how-to/create-initial-data-migrations.md).
- the last line in the migration script should have 4 more spaces of indentation

fyliu:
Fixes #313

code fixes for migration script
- fixed `operations` line indentation
- removed `initial = True`
- added reverse function

- broken code in the guide creates bugs in real code
- `initial = True` is for initial migration and shouldn't be used here.
- reverse function allows jumping to different migrations without errors.

### Testing suggestions

1. Visual check of the `operations` line indentation
2. Visual check of the `initial = True`
3. The below lines
```bash
# run the application
./scripts/buildrun.sh

# the SDG model table should have data
./scripts/db.sh -c "select * from core_sdg;"

# reverse the latest data migration (SDG
./scripts/migrate.sh data 0002

# the SDG model table should be empty
./scripts/db.sh -c "select * from core_sdg;"
```
Old page with problem circled
<kbd>alt-text</kbd>

New page with changes circled
<kbd>alt-text</kbd>

### DIFF
--- a/app/data/migrations/0001_programarea_seed.py
+++ b/app/data/migrations/0001_programarea_seed.py
@@ -3,7 +3,7 @@ from django.db import migrations
 from core.models import ProgramArea
 
 
-def run(__code__, __reverse_code__):
+def forward(__code__, __reverse_code__):
     items = [
         (1, "Citizen Engagement"),
         (2, "Civic Tech Infrastructure"),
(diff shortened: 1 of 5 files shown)


ethanstrominger (changes requested):
Great changes, good notes about how to test.

Optional but a nice to have: add the information in how to test to documentation somewhere - could be part of the create-initial-data-migration script or could be an appendix.  I notice db.sh is not documented anywhere.  Information on how all the scripts work in an appendix would be good.  FYI, the screenshots in the PR were not necessary as I could see the changes in github and your explanation was great.  They look like a lot of work.

If you want to do this later, either you or I could create an issue for this improvement.

ethanstrominger (approved):

PR #342, opened 2024-07-01
Implement user_permissions model, api, and test

### LINKED ISSUE #22: Create Table: user permission
### Dependency

 - #148

### Overview

We need to create the user permission table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items
- [x] identify if table has a description (see spreadsheet under Resources)
  - [ ] if not, reach out to PD leads
- [x] identify and document (below) what the **data fields** are and make sure the ERD and spreadsheets match (see Resources)
- [x] identify and document  (below) what **other tables are associated** and make sure the ERD and spreadsheets match (see Resources)
- [ ] create a single model in Django (defining schema)
- [ ] Write a test for the relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project)
- [ ] Write an API end point
- [ ] write API unit tests
- [ ] Document the endpoint in Swagger

### Resources/Instructions

#### Description
- User permissions for project, brigade, etc.

#### Data Fields
1. Copied from spreadsheet and checked off according to ERD. Data fields are bolded and relation fields are not. (unchecked items indicate a mismatch between ERD and spreadsheet, which requires a review)
   - [x] PK - id - int - Record Id
   - [x] FK - user_id - int - Record Id
   - [x] FK - project_id - int - Record Id
   - [x] FK - permission_type_id - int
   - [x] FK - practice_area_id - int - Record Id
   - [x] granted - timestamp
   - [x] ended - timestamp

   - created_by - int
   - updated_by - int

#### Associated Tables
1. Copied from spreadsheet and checked off according to ERD (unchecked items indicate a mismatch between ERD and spreadsheet, which requires a review)
   - [x] permission_type (many-to-one)
   - [x] practice_area (many-to-one)
   - [x] project (many-to-one)
   - [x] practice_area_id (many-to-one)
   - [x] user (many-to-one)

   - None

#### Swagger Endpoint Link
-

ethanstrominger:
Fixes #22

- **standard changes**:
   - models.py
   - serializers.py
   - views.py
   - urls.py
   - admin.py
   - 00025_user_permissions_and_more.py
- **constants.py**: new file for holding constants for code for practice area admin and project lead.  This will be utilized more going forward.
- **0004_perimission_types_seed.py**: modified to populate rank.  Any databases that applied this will need to reverse it.  I will need this field in another pull request, better to populate it now.
- conftest.py.  Added fixtures: user_superuser_admin, user_permissions (2 generic user permissions), user_permissions_admin_project, user_permissions_practice_lead_project
- test_api.py
   - fixed test_create_permission_type which would otherwise create a duplicate record
   - added test_get_user_permissions.  Corrected sincelast review to get the count from UserPermissions for better stability.
   - added test_user_permission_admin_project and test_user_permission_practice_lead_project to check the str functions under different scenarios


fyliu (changes requested):
Thanks for adding this and making the name more clear. I started the review but keep thinking I should look at it more. But here's what I have so far.

The migration files can be deleted and regenerated as one.

constants.py is only used in one file and seems unnecessary, but I assume it'll have more uses in future PRs.

Maybe add a model test for the str function for coverage to exercise the functionality.

app/constants.py
```diff
@@ -0,0 +1,5 @@
+global_admin = "Global Admin"
```
ExperimentsInHonesty, line 1: Here are the defined names on the PermissionType - Data spreadsheet
https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?pli=1&gid=1448710921#gid=1448710921

adminGlobal
adminVrms
adminBrigade
adminProject
practiceLeadProject
memberProject
memberGeneral

ExperimentsInHonesty (dismissed):
Here are the defined names on the [PermissionType - Data spreadsheet](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?pli=1&gid=1448710921#gid=1448710921) for the constants.py file

adminGlobal
adminVrms
adminBrigade
adminProject
practiceLeadProject
memberProject
memberGeneral

scripts/run.sh
fyliu, line 1: These returns should really be exits. They're only behaving the same because I added the "set -e" guard to the top of the script (all the scripts).

If you removed the "set -euo pipefail" line and ran `./scripts/run.sh -f`, you'll see that it returns for the unrecognized option and then starts the docker containers anyway. It should really exit the entire script instead. Same withh `./scripts/run.sh -u` to print usage

Related change in another script: The returns in `start-local.sh` are okay since they're in the top level scope of the script. They end up behaving the same as exits (except that one return in the function scope that actually returns the value)

fyliu (changes requested):
Can you discard the changes in `run.sh`? I commented on it but it wasn't a review so it didn't show requested changes.

app/core/tests/test_api.py
```diff
@@ -321,6 +322,16 @@ def test_create_stack_element_type(auth_client):
     assert res.data["name"] == payload["name"]
 
 
+def test_get_user_permissions(
+    created_user_admin, created_user_permissions, auth_client
```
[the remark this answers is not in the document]
ethanstrominger: Modified

.pre-commit-config.yaml
fyliu, line 1: This was was fine without the extra lines. They were removed a few commits ago in `main`

app/Dockerfile
```diff
@@ -14,14 +14,16 @@ RUN mkdir -p $PIP_CACHE_DIR
...
+# hadolint ignore=DL3008
 RUN \
   --mount=type=cache,target=/var/cache/apt,sharing=locked \
   --mount=type=cache,target=/var/lib/apt,sharing=locked \
   apt-get update \
   && apt-get install --no-install-recommends -yqq \
   netcat=1.10-46 \
   gcc=4:10.2.1-1 \
-  graphviz=2.42.2-5
+  graphviz
```
fyliu, line 26: Just commenting. @dmartin4820 had the better idea of 2.42.* version. But this is fine since we're about to move to alpine base.

app/core/api/serializers.py
```diff
@@ -38,6 +39,22 @@ class Meta:
...
+        model = UserPermission
+        fields = (
+            "uuid",
+            "created_at",
+            "updated_at",
+            "user",
+            "permission_type",
+            "project",
+            "practice_area",
+        )
```
fyliu, line 55: This is missing the read_only fields that the other models in the same file have. Actually, I'm not sure if they're settable from the API if you didn't declare them `read_only`.

app/core/models.py
```diff
@@ -311,6 +312,36 @@ def __str__(self):
...
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["user", "permission_type", "project", "practice_area"],
+                name="unique_user_permission",
+            )
+        ]
+
+    def __str__(self):
```
fyliu, line 335: Just commenting. The way this is implemented is more like a `__repr__` function than a `__str__` function.

fyliu (changes requested):
Thanks for making the changes. I haven't looked through the test yet, but I wanted to add what I found so far. Some of the comments don't require any change.

fyliu (approved):
Thanks for making the changes!

It's working well and all the tests pass!

PR #376, opened 2024-09-05
Create table project_program_area_xref

### LINKED ISSUE #45: Create Table: project_program_area_xref
### Overview

We need to create the project_program_area_xref table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items

- [x] write an API end point
- [x] write API unit tests
- [x] document the endpoint

### Resources/Instructions

  - ERD
  - Table and Field Definitions
  - API Endpoint

### Description

Join table: links project to program area

### Data Fields

   - [x] (FK) project_id - int - id of project linked to program area
   - [x] (FK) program_area_id - int - id of program area linked to project
   - [x] created - timestamp

   - created_date - timestamp

### Associated Tables

   - [x] program_area (many-to-one)
   - [x] project (many-to-one)

   - None

del9ra:
Fixes #45 

- Created a model in Django
- Wrote a test for the relationships this model will have with other models
- Registered the model in admin.py
- Wrote an API end point
- Wrote API unit tests
- Documented the endpoint

- to create the project_program_area_xref table
- to update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project

No changes


app/Dockerfile
```diff
@@ -20,11 +20,11 @@ RUN \
   apt-get update \
   && apt-get install --no-install-recommends -yqq \
   netcat=1.10-46 \
-  gcc=4:10.2.1-1 \
-  graphviz=2.42.2-5
```
dmartin4820, line 24: To fix the issue we were encountering when running ./scripts/buildrun.sh, removing this line was just a local quick fix. Instead it may be better to specify graphviz=2.42.* so that the newest patch is installed instead (i.e. graphviz_2.42.2-5+deb11u1 instead of graphviz_2.42.2-5).

app/core/models.py
```diff
@@ -383,3 +383,12 @@ class CheckType(AbstractBaseModel):
 
     def __str__(self):
         return f"{self.name}"
+
+
+class ProjectProgramAreaXref(AbstractBaseModel):
+    project_id = models.ForeignKey(Project, on_delete=models.CASCADE)
+    program_area_id = models.ForeignKey(ProgramArea, on_delete=models.CASCADE)
+    created_date = models.DateTimeField("Created at", null=True, blank=True)
```
dmartin4820, line 391: The inherited AbstractBaseModel class already creates a timestamp for creation; see [here](https://github.com/del9ra/peopledepot/blob/9465c5100a9e45c05ad72273fcf22e6839b415a1/app/core/models.py#L12)

dmartin4820 (changes requested):
Minor changes to version in Dockerfile and duplicate created timestamp field in model. 

Also, the past pull requests I've looked at will merge into hackforla:main, unless @fyliu or someone else said different. My last create table PR was directly into hackforla:main.

app/core/models.py
del9ra: Why did they use the DateTimeField in the Add the Model section of this guide: https://hackforla.github.io/peopledepot/how-to/add-model-and-api-endpoints/, and in the Project model as "completed_at", in the Event model as "start_time", and in the Affiliation model as "ended_at"?

del9ra:
> Minor changes to version in Dockerfile and duplicate created timestamp field in model.
> 
> Also, the past pull requests I've looked at will merge into hackforla:main, unless @fyliu or someone else said different. My last create table PR was directly into hackforla:main.

@fyliu Could you clarify this please?

app/core/models.py
fyliu: @del9ra there's some translation required between the DB notation in the ERD and spreadsheet and the Django data fields. We use an abstract base class (ABC) so that all the models will have create and modify timestamps. In the database design, many but not all models have the timestamps. We decided to make that a standard thing at one point.

So you should ignore anything from the database table that does the same as create and update timestamps.

At the time, I was trying to decide what to call these standard timestamps, and I found some page which discussed "on" vs "at", where "on" is usually a date and "at" is usually a time. So I added "_at" to the end of the names. But eventually I found a similar `TimeStampedModel` ABC in the widely-used django package [django-extensions](https://django-extensions.readthedocs.io/en/latest/model_extensions.html#database-model-extensions). It would've been a good idea to use that too.

fyliu:
@del9ra He means the `to` branch should be `hackforla:main`. Your `from` is good, but your `to` is set to another branch in `hackforla`. It should be from your feature branch in your fork to `hackforla:main`.

Here are screenshots from this and Denzal's PR

2024-09-08 01 52 57 github com 0668c51bf5e9
2024-09-08 01 54 44 github com bab2b93fdd5c

app/core/models.py
fyliu: Hmm... looks like this xref table doesn't have any extra data fields other than the relations (foreign keys). It means that we don't really need to create a Django model for this table at all. I'll have to go though the other issues and pick out similar issues before someone works on them.

fyliu: @del9ra I'm sorry, but I need to change this issue from "create xref table" to "create many to many relation between Project and ProgramArea models". We don't have any existing code in the repository to serve as an example, but I can find some other examples if you want to work on it.

dmartin4820: > Why did they use the DateTimeField in the Add the Model section of this guide: https://hackforla.github.io/peopledepot/how-to/add-model-and-api-endpoints/, and in the Project model as "completed_at", in the Event model as "start_time", and in the Affiliation model as "ended_at"?

@del9ra Did you get your question answered? I'm not sure why those were included either. This should be added to the meeting agenda if we can't resolve this. I'm not sure of how the decisions were made for completed, start, ended and so on. Looking at the ERD also shows that completed_at has a type of "date" versus "DateTimeField". It looks like the ERD also needs to be updated for at least Project, if not all the other tables.

dmartin4820:
> @del9ra He means the `to` branch should be `hackforla:main`. Your `from` is good, but your `to` is set to another branch in `hackforla`. It should be from your feature branch in your fork to `hackforla:main`.
> 
> Here are screenshots from this and Denzal's PR
> 
> 2024-09-08 01 52 57 github com 0668c51bf5e9 2024-09-08 01 54 44 github com bab2b93fdd5c

If we want this to merge into main, I don't know the best/easiest way to do that since this PR has already been created. It may be possible to let it merge into ```[hackforla:45-create-table-project_program_area_xref```, then merge that branch into main. The other way is to just create a new PR. I'm not sure there is a clean way to change the PR branch destination at this point, but the changes will still be able to go to the correct place regardless.

shmonks:
Closing this PR because it turns out that most xref tables can be created automatically by Django

app/core/models.py
del9ra: > @del9ra there's some translation required between the DB notation in the ERD and spreadsheet and the Django data fields. We use an abstract base class (ABC) so that all the models will have create and modify timestamps. In the database design, many but not all models have the timestamps. We decided to make that a standard thing at one point.
> 
> So you should ignore anything from the database table that does the same as create and update timestamps.
> 
> At the time, I was trying to decide what to call these standard timestamps, and I found some page which discussed "on" vs "at", where "on" is usually a date and "at" is usually a time. So I added "_at" to the end of the names. But eventually I found a similar `TimeStampedModel` ABC in the widely-used django package [django-extensions](https://django-extensions.readthedocs.io/en/latest/model_extensions.html#database-model-extensions). It would've been a good idea to use that too.

@fyliu I noticed that we have the `ended_at` field in the Affiliation model; `ended` in Permission model, project_sponsor_partner_xref, permission_history; `ended_on` in project_language_xref and project_sdg_xref. How does this work in this case? Do we need to create a single field name for them in the ABC model?

fyliu: @del9ra we could. It looks like the `ended` field applies to xref tables where the relationships ended but we're keeping the row around (soft delete). But we don't have a lot of those tables to apply it to, so we shouldn't do this for all the models. Then again, you pointed out that the few are already inconsistent in naming, so maybe we should do something to make/keep them consistent.

I'm thinking we should bring up the inconsistent naming to @ExperimentsInHonesty and @Neecolaa and at least make an issue (ER maybe) to fix it on some tables. 

We should bring this to the meeting and decide on whether or not to make an ABC model or something else to help force this consistency.

PR #404, opened 2024-10-08
Make project optional 

### LINKED ISSUE #403: Make project optional for user_permissions for global admin
### Overview
Enable admin to create a global admin that can see and update all records, regardless of project.

ethanstrominger:
Fixes #403 

Changed models.py so project was optional for user_permissions and created migrate scripts.


dmartin4820 (approved):
Looks good to me

PR #406, opened 2024-10-08
Prep work for #401

### LINKED ISSUE #405: Row and field level security prep work
### Overview

Ground work that needs to be done to support issue  #401.

ethanstrominger:
Fixes #405


.pre-commit-config.yaml
```diff
@@ -32,7 +32,7 @@ repos:
       - id: fix-byte-order-marker
       - id: name-tests-test
         args: [--pytest-test-first]
-
+        exclude: ^app/core/tests/utils/
```
dmartin4820, line 35: It looks like files currently don't exist in this directory, but files will be added in PR #401. How do we decide that these files are excluded? I have no issue with this ignore, but I'm not sure about the reason. Don't we want it to check those files too?

dmartin4820 (approved):
Just one comment, but more for clarification rather than a fix. Otherwise it looks fine to me.

I also ran with Ethan's changes and had no issues with the admin page, building, or the tests.

PR #449, opened 2024-11-21
Create table project status 46

### LINKED ISSUE #46: Create Table: project_status (used to say project_program_area_status)
### Overview

We need to create the project_status table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items

- [x] ~~Create issue for updating `project`table connecting `current_status_id` to `project_status`~~
-- This was done in the PR attached to this issue instead of creating a new one. 
- [x] write an API end point
- [x] write API unit tests
- [x] document the endpoint

### Post issue tasks

- [ ] resolve dependency on #67 and move it from Ice Box to New Issue Review with `ready for tech lead` label

### Resources/Instructions

  - ERD
  - Table and Field Definitions
  - API Endpoint

### Description

Dictionary of status options for project and program_area

### Data Fields

   - [x] name - varchar
   - [x] description - varchar

   - None

### Associated Tables

   - [x] project (one-to-many) #16
   - [x] project_status_history (one-to-many) #67 

   - None

dmartin4820:
Fixes #46

- created project_status model
- register project_status in admin
- add project_status endpoints
- add project_status seed data

N/A


del9ra (changes requested):
Hi Denzel,
Thank you for working on this! I tested it, and everything worked well on my end. I just have a couple of questions:
1. Why are the descriptions on the 5th and 6th lines in quotation marks in the app/core/initial_data/PD_Table and Field Explanations - ProjectStatus - Data.csv file? Do you think this has an impact on how the code works?
2. Shouldn't the uuid field be mentioned in the _seed.py file?

dmartin4820:
> Hi Denzel, Thank you for working on this! I tested it, and everything worked well on my end. I just have a couple of questions:
> 
>     1. Why are the descriptions on the 5th and 6th lines in quotation marks in the app/core/initial_data/PD_Table and Field Explanations - ProjectStatus - Data.csv file? Do you think this has an impact on how the code works?
> 
>     2. Shouldn't the uuid field be mentioned in the _seed.py file?

Hey Del, thanks for the comments.

1. I'm unsure how that got that way. The excel sheet doesn't have the quotes. I'll change that.
2. Good point. All the other seed files have UUID, but the [table in the excel sheet,](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=419535315#gid=419535315) for this PR, didn't have an id column

fyliu:
> 2. Good point. All the other seed files have UUID, but the [table in the Google sheet,](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=419535315#gid=419535315) for this PR, didn't have an UUID column. @Neecolaa @fyliu Can we make an update to[ the Project Status table in Google sheets](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=419535315#gid=419535315) so that it contains an id column? By doing this we would maintain consistency with the other tables and seed data with UUID fields.

I guess it depends on if we still need it set to specific values. The values it's being set to in the spreadsheet aren't really UUID values.

Originally, I let the database generate the ID/UUID values since it's an auto-generated field and the initial data sheets didn't have the values. Then @ethanstrominger said he needed to set the values for testing purposes, so we did that for some of the tables. But I don't think it's being used in tests right now?

We should bring this up at the meeting and ask if we still need to set this.

del9ra (approved):
Hi @dmartin4820 ,
Thanks for your contribution! Nice work on this, it looks great to me!

app/core/models.py
```diff
@@ -137,7 +149,9 @@ class Project(AbstractBaseModel):
 "Authorization: token [gh_PAT]" \
 https://api.github.com/repos/[org]/[repo]',
     )
-    # current_status_id = models.ForeignKey("status", on_delete=models.PROTECT)
+    current_status_id = models.ForeignKey(
```
fyliu, line 152: This should be `current_status`.

The old code was written before we realized Django convention is to drop the `_id` ending for foreign key relations.

app/core/initial_data/PD_ Table and field explanations - ProjectStatus - Data.csv
```diff
@@ -0,0 +1,6 @@
+name,description
+Active,Has a project team and current meetings
+On Hold,No project team or meetings scheduled
+Completed,Project is completed
+Closed,Closed, possibly not completed (usually does not show up on the website).  Unlikely to be reopened.
+Deleted,Holds for 90 days before final removal (used for test project entries, or mistakes that do not need to be remembered)
```
fyliu, line 6: Sorry, can you add back the double quotes?

I think the double quotes are actually necessary for the convert.py script to work. They're likely added by google to help any csv parser read it.

fyliu (changes requested):
Thanks for working on this! I looked at this more closely and found a couple more things to improve.

app/core/initial_data/PD_ Table and field explanations - ProjectStatus - Data.csv
dmartin4820: Added back. It was definitely needed otherwise the csv parser would not run.

fyliu (changes requested):
Thanks for making the changes. They look great!

I'm sorry, I just noticed another little thing.

The admin site adds an "s" at the end when displaying the plural form of the model. It doesn't know to use "project statuses". 
2025-02-27 15 26 00 localhost 6b88d8307465

In  `models.py` you can search for "verbose_name_plural" and use that as an example to add the plural for it. I think you should make it all lowercase and let Django deal with the capitalization. I see the existing code is doing something different.

dmartin4820:
> Thanks for making the changes. They look great!
> 
> I'm sorry, I just noticed another little thing.
> 
> The admin site adds an "s" at the end when displaying the plural form of the model. It doesn't know to use "project statuses". 2025-02-27 15 26 00 localhost 6b88d8307465
> 
> In `models.py` you can search for "verbose_name_plural" and use that as an example to add the plural for it. I think you should make it all lowercase and let Django deal with the capitalization. I see the existing code is doing something different.
I went and added the line for project status and checked that I could see the correct plural in the admin page:

project_status_plural_change.

Don't know why my page is acting that way, but either way project status is plural as expected. Let me know if there's anything else. Thank you for the review and additions so far.

fyliu (approved):
Thanks. It looks great!

I'll combine the fixes into the original commits and merge them in.

PR #473, opened 2025-02-26
Replaced spaces with commas in ALLOWED_HOSTS

### LINKED ISSUE #469: Use Django(Python)'s convention for ALLOWED_HOST list value delimiter
### Overview

We want to follow Django's convention of using a comma-separated list for the `ALLOWED_HOSTS` configuration value.

### Action Items

- [x] Make the change to use comma list delimiter and modify the example config file
- [x] Warn other developers that they'll need to update their config files manually
  - [ ] at a team meeting
  - [x] on slack
- [ ] close ER #468

### Resources/Instructions

- [Django docs on `ALLOWED_HOSTS`](https://docs.djangoproject.com/en/5.1/ref/settings/#allowed-hosts) showing an example of the list that's comma-separated.

del9ra:
Fixes #469 

- Replaced spaces with commas in `settings.py` and `.env.docker-example `

- To follow Django's convention of using a comma-separated list for the ALLOWED_HOSTS configuration value


fyliu (approved):
Great work!

What I did to test:
1. I made sure the old space-separated list breaks the site, and it did complain about localhost not in ALLOWED_HOSTS.
2. Then I made sure the new comma-separated list made the site work again, and it does.

Issue #68, opened 2022-09-11
Create Table: referrer


fyliu:
### Overview

We need to create the referrer table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

A table and a model are the same thing

### Action Items

- [x] identify and document table description (see spreadsheet under Resources)
  - [ ] if not, reach out to PD leads
- [x] compare and check off the data fields (below) against the ERD. Note any fields in the ERD that's missing from the list. (see Resources)
- [x] compare the check off associated tables (below) against the ERD. Note any associated tables in the ERD that's missing from the list (see Resources)
- [x] create a single model in Django (defining schema)
- [x] write a test for the relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project).
- [x] write an API end point
- [x] write API unit tests
- [x] document the endpoint
- [x] uncomment `referred_by` field in User model when this issue's table is created.
- Check to see if open issues in `FK status` below
   - [ ] If all Issues listed are closed (or there are no FKs), then skip the next 3 steps
      - [x] comment out the code after you create it (see FK Status below for open/closed state)
      - [x] add an action item on the open issue to uncomment the code line you commented out when that issue's table is created. 
      - [x] Provide your file and line number as a permalink under resources in that issue

### Resources/Instructions

- See [People Depot Resources wiki page](https://github.com/hackforla/peopledepot/wiki/Resources-and-Links) for links
  - ERD
  - Table and Field Definitions
  - API Endpoint
- See the [wiki page](https://github.com/hackforla/peopledepot/wiki/Create-Table-issues-data-gathering-workflow) for instructions.

### Description

List of referrers

### Data Fields

Name | Type | Required | FK Table | FK table Issue(s)
-- | -- | -- | -- | --
name | varchar() | TRUE | | 
url | url varchar |  | | 
referrer_type_id | int | TRUE | referrer_type | #69
contact_name | varchar |  | | 
contact_email | email varchar |  | | 

1. In ERD only (having items here indicates a mismatch, which requires a review)
   - None
   - 
### FK status
- #69 

### Associated Tables

1. Copied from spreadsheet and checked off according to ERD. (unchecked items indicate a mismatch between ERD and spreadsheet, which requires a review)
   - [x] user (one-to-many)
   - [x] referrer_type (many-to-one) #69 

1. In ERD only (having items here indicates a mismatch, which requires a review)
   - None

shmonks:
[...]

PR #477, opened 2025-03-09
Updated User model variable names

### LINKED ISSUE #429: Update Table: User
### Overview
We need to update the User table model to include new fields and have the correct names for fields.

#### Details
- The initial model issue: #15

### Action Items
Update existing Django model
- [x] Uncomment the referred_by field in the User model once the referrer table is created: https://github.com/hackforla/peopledepot/blob/b97453e67147c4bf1ccc210ff50d76282df812b5/app/core/models.py#L90
- [x] In the files indicated by Resource 1.01, Change the following items in the code 

Current name in code | Updated Name | Updated Type (may already be this type) | FK Table | FK Issues
-- | -- | -- | -- | --
current_job_title | job_title_current_intake | varchar | | 
~~email~~ | ~~email_intake~~ | EmailField | | 
~~first_name~~ | ~~name_first~~ | varchar | | 
gmail | email_gmail | EmailField | | 
~~is_active~~ | user_status_id | int | user_status_type | #35
~~last_name~~ | ~~name_last~~ | varchar | | 
preferred_email | email_preferred | EmailField | | 
target_job_title | job_title_target_intake | varchar | | 

- [x] Add the following items in the code
- [ ] 

Name | Type | FK Table | FK table Issue(s)
-- | -- | -- | --
practice_area_primary | int (practice_area_primary) | practice_area | #63
practice_area_secondary | int[] (practice_area_secondary) | practice_area | #63
practice_area_target_intake | int[] (practice_area_target_intake) | practice_area | #63
~~referred_by~~ referrer | int | referrer | #68
email_cognito | EmailField | | 

- [x] Write a test for the new relationships this model will have with other models (e.g., creating a user and assigning them a set of permissions on a project) if any.
- [x] Update API end point
- [x] Update API unit tests
- [x] [generate a schema table definition](https://github.com/hackforla/peopledepot/issues/378#issuecomment-2493062821) and post it in a comment below and add a link to it in the Update Schema dependency issue.
- [x] For each of the FK Above if the FK table has not yet been created (i.e, the issue is still open), you will need to
   - comment out the code after you create it 
   - add an action item on the open issue to uncomment the code line you commented out when that issue's table is created. 
   - Provide your file and line number as a permalink under resources in that issue
- [ ] Document the endpoint in ReDocs

### After PR has been approved
- [ ] Release the dependency and move the issue out of the icebox for #436

### Resources
- 1.01 Code locations
   - 1.01.01 [/app/core/models.py](https://github.com/hackforla/peopledepot/blob/main/app/core/models.py)
   - 1.01.02 [/app/core/admin.py](https://github.com/hackforla/peopledepot/blob/main/app/core/admin.py)
   - 1.01.03 [/app/core/api/serializers.py](https://github.com/hackforla/peopledepot/blob/main/app/core/api/serializers.py)
   - 1.01.04 [/app/core/migrations/0002_user_current_job_title_user_current_skills_and_more.py](https://github.com/hackforla/peopledepot/blob/main/app/core/migrations/0002_user_current_job_title_user_current_skills_and_more.py)
- 1.02 [People Depot Resources wiki page](https://github.com/hackforla/peopledepot/wiki/Resources-and-Links) for links
  - ERD
  - Table and Field Definitions
  - API Endpoint
- 1.03 [Spreadsheet, User view filter](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=371053454#gid=371053454&fvid=139862014)

#### FK status
- #35
- #63 
- #68

del9ra:
Fixes #429 

- Updated  names and added the new items in views.py, conftest.py, test_models.py, test_apis.py, admin.py, serializers.py and models.py
- Added a new UserManager model that doesn't use `email` because of the issues with `email` field
- Wrote a test for the new relationships this model has with `UserStatusType` and `PracticeArea`
- Wrote a test to check that the old names will return errors
- Added an action item to issue #68   to uncomment the code line we commented out when that issue's table is created.
- Generated a schema table for User, post it in a comment below and added a link to it in the Update Schema dependency issue.

- to include new fields and have the correct names for fields


shmonks:
Just converting this to draft for now, @del9ra

del9ra:
> Just converting this to draft for now, @del9ra

Understood, thanks!

ExperimentsInHonesty:
Need to review with @Neecolaa the user table... this PR has not merged yet.  See [PD: Table and field explanations spreadsheet with filter user](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=371053454#gid=371053454&fvid=2121927878)

ExperimentsInHonesty:
@del9ra the referred_by field by default should be null unless the user provides other information.  Once you fix that, please resubmit for review

app/core/models.py
```diff
@@ -62,26 +105,37 @@ class User(PermissionsMixin, AbstractBaseUser, AbstractBaseModel):
...
+    email_gmail = models.EmailField(blank=True)
+    email_preferred = models.EmailField(blank=True)
+    email_cognito = models.EmailField(blank=True)
 
     user_status = models.ForeignKey(
         "UserStatusType", null=True, on_delete=models.PROTECT
     )
     # current_practice_area = models.ManyToManyField("PracticeArea")
-    # target_practice_area = models.ManyToManyField("PracticeArea")
+    practice_area_primary = models.ManyToManyField(
```
fyliu, line 127: I think this field is just a FK relation rather than many-to-many.

Also, you should delete the "current_practice_area" comment above as well.

fyliu:
I'm getting this error and having second thoughts about changing the `email` field to `email_intake`. 

Here's a message that our change somehow caused the `email_intake` to be required for the `createsuperuser` command.
image


It adds more maintenance work beyond the simple name change. Django and some Django plugins are depending on the `email` name by default. It can be changed, but we would have to do the customizations when we encounter them. It would be less work if we kept the Django default field names for `email` and a few others.

The Django default fields are these (from AI, bolding names that might change):

1. `id`: An auto-incrementing primary key field that uniquely identifies each user.
2. `password`: A field that stores the user's password, which is hashed for security.
3. `last_login`: A field that stores the date and time of the user's last login.
4. `is_superuser`: A boolean field that indicates whether the user is a superuser (i.e., has all permissions).
5. `username`: A field that stores the user's username, which must be unique.
6. **`first_name`**: A field that stores the user's first name.
7. **`last_name`**: A field that stores the user's last name.
8. **`email`**: A field that stores the user's email address.
9. `is_staff`: A boolean field that indicates whether the user is a staff member (i.e., has access to the admin interface).
10. **`is_active`**: A boolean field that indicates whether the user's account is active.
11. `date_joined`: A field that stores the date and time when the user's account was created.

fyliu (changes requested):
Just pointing out a couple things that needs changing:
1. `referrer` -> `referred_by` field
   - it's supposed to be a new field in the issue description, but it's implemented as a different name so it needs to be updated.
3. The `DJANGO_SUPERUSER_EMAIL` in `.env.docker-example` needs to be updated for the `email_intake` name change. Just add `_INTAKE` at the end there.

app/core/tests/test_models.py
```diff
@@ -248,3 +251,83 @@ def test_referrer_has_a_referrer_type(referrer):
...
+    inactive_user_status = UserStatusType.objects.get(name="Inactive")
+    active_user_status.user_set.add(user)
+    active_user_status.user_set.add(user2)
+
+    assert active_user_status.user_set.count() == 2
+
+    assert user.user_status == active_user_status
+    assert user2.user_status == active_user_status
+
+    active_user_status.user_set.remove(user)
```
fyliu, line 289: You don't need to remove one status before adding another one. You can just set the user's status to inactive or add the user to the inactive_user_status user_set in the way you're writing it.

fyliu:
I think this PR is great other than the few minor things I mentioned.

The only thing I'm concerned about right now is the removal of the `is_active` field.
- **Background**: Django comes with code to read that field when users log in, and reject it if a user is not active. This allows the system to retain the user's records while disallowing login, like a soft delete. 
- **Effect of removing it**: Without `is_active`, there's no way to bar users from logging in, other than deleting the user's record from the `user` table.
- **Our plan**: We're going to replace this with some `user_status` lookup logic, but we don't have that working yet. So I want to make sure we really want to remove `is_active` right now.
- **Suggestion**: An alternative I'm think of is we can mark the field as "deprecated" (in the model/table field description), and hide it from the API for now. Then, when we do implement the replacement logic, we can remove the field.
   - The developer for that issue would need to look up all the places that reference `is_active` and override them with the new logic. I mean that'd be a required part of that issue anyway. It might even be helpful to have the `is_active` field around so that the developer doesn't have to "know" what name to search for.

fyliu:
Ah, the spreadsheet says the field should be called `referrer`. So everything in the PR looks good.

PR #481, opened 2025-03-10
480 add a simple page at the site root

### LINKED ISSUE #480: Add a blank homepage to the backend
### Overview

The backend needs to return a page at the root url "/" or else the load balancer will spam the logs repeatedly forever.

This is a quick fix and the code is already done.

### Action Items

- [ ] return a 200 status code with a blank page at "/"

fyliu:
The purpose of this is to have a valid page at / to appease the load-balancer so it doesn't spam the logs.

Fixes #480

- added a simple page at "/"

- so the load balancer will stop spamming the logs with errors

#### Before

<kbd>2025-03-10 15 11 26 peopledepot-dev vrms io f4bb6cc7a7de</kbd>

#### After

<kbd>2025-03-10 15 08 09 localhost b3c3460c5035</kbd>

### DIFF
--- a/app/peopledepot/urls.py
+++ b/app/peopledepot/urls.py
@@ -1,11 +1,18 @@
 from django.contrib import admin
+from django.http import HttpResponse
 from django.urls import include
 from django.urls import path
 from drf_spectacular.views import SpectacularAPIView
 from drf_spectacular.views import SpectacularRedocView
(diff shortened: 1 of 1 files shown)


dmartin4820 (approved):

dmartin4820:
I checked with and without the change and see that the expected result on the root page.

PR #491, opened 2025-05-05
Create table leadership type 42

### LINKED ISSUE #42: Create Table: leadership_type :smile:
### Overview

We need to create the leadership_type table so that we can update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.

#### Details

### Action Items

- [x] add seed data
- [x] write an API end point
- [x] write API unit tests
- [x] document the endpoint

### Resources/Instructions

  - ERD
  - Table and Field Definitions
  - API Endpoint

### Description

Dictionary of leadership types to be associated with a project i.e. "Mentor Led", "Peer Led"

### Data Fields

1. Copied from spreadsheet and checked off according to ERD. 

 Name | Type 
-- | --
name | varchar()
description | varchar() 

   - None

### Associated Tables

   - [ ] project (one-to-many)

   - None

dmartin4820:
Fixes #42 

- create leadership_type model
- register leadership_type in admin
- add leadership_type endpoints
- add leadership_type seed data
- add tests for leadership_type creation
- add test for leadership_type to project relationship

- [Here](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=1554600661#gid=1554600661) the seed data has an id column with blank id fields. This creates an empty column in the resulting .csv file used for creating seed data. To complete adding the seed data, I manually removed this column from the .csv file after dowloading from the Google sheet.
- Uncommented the leadership_type foreign key [here](https://github.com/hackforla/peopledepot/blob/b97453e67147c4bf1ccc210ff50d76282df812b5/app/core/models.py#L162). This is done so that the relationship is testable.


del9ra (changes requested):

del9ra (approved):
Thank you, @dmartin4820 !!

app/core/tests/test_api.py
fyliu, line 1: Can you also test the project-leadership_type relationship by assigning a project's leadership_type?

You can do something like this to update an individual field in a project
`res = auth_client.patch(PROJECTS_URL + str(project.pk) + "/", payload)`

fyliu (changes requested):
It looks great!

Can you also test the relationship in the API just to make sure it'll work  for the client using it?

app/core/tests/test_api.py
dmartin4820: The change suggested was made. There wasn't documentation on it so far, so I had to cobble up some examples and code from Django's docs. I'm open to more suggestions if there is a better way to do things. Thanks for the hints and comments.

fyliu (changes requested):
Sorry, I couldn't quite figure out what I didn't like about the solution you came up with. Now, I realize it's because it's we don't really need the extra endpoint. The existing endpoint works well for the same purpose.

I think what you found is a useful solution for when we need to check in users, maybe something like `/users/<pk>/events/<pk>/check-in`.

It's been a while, so I can make the changes if you don't want to get back into this code.

fyliu (approved):
Thanks for making the changes! I reorganized some of the commits for adding to `main`. The overall changes are exactly the same.

PR #492, opened 2025-05-18
Create table referrer 68

### LINKED ISSUE #68: Create Table: referrer

del9ra:
Fixes #68 

- created a Referrer model
- wrote tests
- wrote an API end point and API unit test
- documented the endpoint

- update a shared data store across hackforla.org, vrms, civictechjobs, and tables (onboarding) project.


app/core/api/serializers.py
```diff
@@ -431,3 +432,12 @@ class Meta:
...
         fields = ("uuid", "name", "description")
         read_only_fields = ("uuid", "created_at", "updated_at")
+
+
+class ReferrerSerializer(serializers.ModelSerializer):
+    """Used to retrieve referrer info"""
+
+    class Meta:
+        model = Referrer
+        fields = ("uuid", "name")
```
dmartin4820, line 442: Should we also add, at least, URL and referrer_type to the fields tuple? Adding contact_name and contact_email is debatable based on whether that is too much information.

dmartin4820 (changes requested):
Just left one comment on the PR (serializers.py). 

One thing I noticed was that this model didn't have seed data in the [Google sheet](https://docs.google.com/spreadsheets/d/1x_zZ8JLS2hO-zG0jUocOJmX16jh-DF5dccrd_OEGNZ0/edit?gid=1572339087#gid=1572339087). This may not necessarily have to be in the PR, but it should be noted somewhere (possibly "emergent issue") in case it becomes relevant in the future.

dmartin4820 (approved):

app/core/api/serializers.py
del9ra: @dmartin4820 I've opened an ER for adding the Referrer seed data to the Google Sheet: https://github.com/hackforla/peopledepot/issues/493

fyliu (changes requested):
The new model seems to work fine. It's just missing a FK field from the User model.

fyliu (changes requested):
This is almost all working.

The only thing is there's no way to create a referrer through the API and supply the `contact_name` and `contact_email`.

Related to this but not something you need to do, is that the API doesn't enforce the required constraint of the model. We can look at newer Django or DRF versions to see if they added that.

Issue #490, opened 2025-04-21
PD: PM/DB Lead


ExperimentsInHonesty:
### Overview

Gather all items team members want to propose for this week's agenda.

### Action Items

Add a comment to this issue with your proposed agenda item. If item warrants an issue of its own, please start a new issue, by going to the issues tab above and clicking on the New Issues green button.

Issue Template
```
## [YYYY-MM-DD] Meeting Agenda


### New Items
- [ ] 
- [ ] Carry over Items to next meeting that don't get addressed

### FYIs
-

### Notes from Meeting

### Task Items
```

### Resources
If we come across an issue that needs to be reformatted with a table or FK info, etc.  Use the templates in these comments to address
- https://github.com/hackforla/peopledepot/issues/472

fyliu:
We talked with Ethan about security issues such as Single sign-on and permissions

Todo:

- Fang:

  - [ ] create issue for Ethan to commit his React client code #533
  - [ ] point out in PR #346 what is confusing so Ethan can try to fix the same type of problems

- Ethan:

  - [ ] review and commit React client code to https://github.com/hackforla/peopledepot-react-sample
  - [ ] go through permissions PR #346 and fix anything confusing to review

ExperimentsInHonesty:
I tried talking to AWS from that account
I got an email 
Hello from Amazon Web Services,

Account number: 853744273466

Your AWS management or payer account 972670934240 has a past due balance that requires immediate attention. If we do not receive payment, your account may be restricted or suspended.

To avoid restrictions or suspension, please contact your representative responsible for making payments to AWS and pay the past due charges.
4:03 PM


Please let me look into your accout deeply 
4:03 PM

If your account becomes restricted, you will experience limited ability to buy new resources, expand usage, or link new accounts. If your account is suspended, you will not be able to access your AWS Management Console, manage existing resources, or buy any new resources.

If you want to use and pay for AWS services independently, you can learn how to end your relationship with your organization’s management account.

If you have any questions, contact AWS Customer Service.

Thank you,
Amazon Web Services
4:03 PM

it looks like the account got suspended, but when I looked inside it I didn't see any way to see the amount overdue or who the organizational owner is, 
4:04 PM


Please allow me to continue working on your case i will be right back
4:06 PM

sounds good
4:06 PM


I apologize for taking this time long, only the management account can discuss account or billing-related details with you. For help with this issue, contact your management account. You can find your management account's details here:

https://console.aws.amazon.com/organizations/home
[...]

PR #506, opened 2025-08-11
Adding schema erd document

Neecolaa:
- Not fixing a specific issue, discussed in #490 

- Added the file `schema.dbml` which contains our current ERD

- We've moved our ERD from Lucidchart to dbdiagram.io so that we can have version control and more members can update the document.
- The history of changes made to schema.dbml will serve as our history for changes made to the ERD


ExperimentsInHonesty (approved):
LGTM

PR #541, opened 2025-09-09
Change docker-compose to docker compose

### LINKED ISSUE #540: Scripts that refer to docker-compose need to switch to docker compose
Starting with 4.32 docker no longer includes docker-compose.  docker compose will work in its stead.  docker compose has been supported for a while so existing devs should be okay.

See https://docs.docker.com/compose/how-tos/networking and https://stackoverflow.com/questions/78769459/command-not-found-docker-compose-after-docker-4-32-0-update

ethanstrominger:
Fixes #540

Changed docker-compose to docker compose.
docker-compose has been desupported


fyliu (approved):
Looks good! Thanks for updating the code!

PR #641, opened 2026-02-16
Update issue templates

ExperimentsInHonesty:
This PR is to update the seed issue template. 

Currently there is an outstanding question though.  

The Template says
>2.02 json file (after commit):

But should that still be a json?  We changed other instances of json to csv.  And where were the person working on this issue find that json file after the commit.  That is not clear.


fyliu (approved):
Looks good. Thanks!

Issue #675, opened 2026-05-16
Update deployment workflow


fyliu:
### Overview

Update the deployment workflow(s) to use the new AWS Docker configuration with gunicorn and Whitenoise instead of the development setup.

### Action Items
- Update `deploy-dev.yml` (or create a new workflow) to:
  - Build using `Dockerfile-aws` (if not already using it)
  - Use `requirements-aws.txt`
  - Call `entrypoint-aws.sh` correctly
  - Ensure environment variables are set for production/AWS
- Verify that the workflow builds and deploys successfully
- Test that the deployed app serves static files correctly with Whitenoise
- Update any documentation that references the deployment process

### Acceptance Criteria
- Deployment workflow builds the Docker image successfully
- Workflow uses the correct Dockerfile and requirements
- App deploys and starts with gunicorn
- Static files are served correctly in the deployed environment
- No breaking changes to the dev environment

**Parent Issue**: #219
**Blocked by**: #219 sub-issue "Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh" (or the multi-stage optimization if you wait for that)

Issue #674, opened 2026-05-16
Create docker-compose-aws.yml for local testing


fyliu:
### Overview

Create a docker-compose file that allows developers to test the AWS/production Docker setup locally before deploying. This helps catch configuration issues early.

### Action Items
- Create `docker-compose-aws.yml` that:
  - Builds using `Dockerfile-aws`
  - Uses `requirements-aws.txt`
  - Runs the `entrypoint-aws.sh` script
  - Starts gunicorn (not runserver)
  - Includes database service for local testing
- Add documentation explaining how to use the file for testing
- Test that the setup works locally without issues

### Acceptance Criteria
- `docker-compose-aws.yml` exists and builds successfully
- App starts with gunicorn and serves static files
- Database migrations run correctly
- Developers can test the production setup locally
- Instructions are documented for developers

**Parent Issue**: #219
**Blocked by**: #219 sub-issue "Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh"

Issue #673, opened 2026-05-16
Optimize Dockerfile-aws with multi-stage build


fyliu:
### Overview

Refactor `Dockerfile-aws` to use a multi-stage build to reduce the final production image size. This removes build-time dependencies and temporary files that don't need to be in the final image.

### Action Items
- [ ] Refactor `app/Dockerfile-aws` to use a multi-stage build:
  - [ ] Stage 1: Builder stage that installs dependencies and runs `collectstatic`
  - [ ] Stage 2: Runtime stage that copies only what's needed from the builder
- [ ] Remove build artifacts and cache from the final image
- [ ] Compare final image sizes before and after optimization
- [ ] Test that the optimized image builds successfully and the app runs correctly
- [ ] Document any image size improvements

### Acceptance Criteria
- Multi-stage build is implemented in `Dockerfile-aws`
- Final image size is smaller than the single-stage version
- App starts and serves static files correctly in the optimized image
- Build process completes without errors
- Image size comparison is documented (in commit message or PR description)

### References
- [Docker multi-stage builds](https://docs.docker.com/build/building/multi-stage/)
- We're not sure how much this will reduce the image size, since we already have method to remove build caches from the image.

**Parent Issue**: #219
**Blocked by**: #219 sub-issue "Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh"

Issue #672, opened 2026-05-16
Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh


fyliu:
### Overview

Set up a production-ready Docker configuration for AWS deployment using gunicorn as the WSGI server and Whitenoise for static file serving. This builds on the split requirements from the previous issue.

### Action Items
- [ ] Add `gunicorn` and `whitenoise` to `app/requirements-aws.in`
- [ ] Recompile `app/requirements-aws.txt`
- [ ] Create or update `app/Dockerfile-aws` to:
  - [ ] Use `requirements-aws.txt` instead of `requirements.txt`
  - [ ] Run `python manage.py collectstatic --noinput` during the build
  - (Use single-stage build for now—optimization comes later)
- [ ] Create or update `app/entrypoint-aws.sh` to:
  - [ ] Run database migrations if needed
  - [ ] Start gunicorn as the WSGI server (e.g., `gunicorn peopledepot.wsgi:application --bind 0.0.0.0:8000`)
- [ ] Update `app/peopledepot/settings.py` to configure Whitenoise:
  - [ ] Add `'whitenoise.middleware.WhiteNoiseMiddleware'` to `MIDDLEWARE` (near the top, after SecurityMiddleware)
  - [ ] Set `STATIC_ROOT` and `STATIC_URL` appropriately
  - [ ] Optionally configure `WHITENOISE_AUTOREFRESH` or other Whitenoise settings
- [ ] Test that the Dockerfile builds successfully and the entrypoint runs without errors

### Acceptance Criteria
- `Dockerfile-aws` builds successfully and uses `requirements-aws.txt`
- `entrypoint-aws.sh` exists and starts gunicorn
- Whitenoise is configured in Django settings
- Static files are collected during the Docker build
- The app starts and serves static files when the container runs
- Single-stage build (no multi-stage optimization yet)

### References
- [Django static files guide](https://testdriven.io/blog/django-static-files/)
- [Whitenoise in Docker Deployment](https://testdriven.io/blog/deploying-django-to-heroku-with-docker/#whitenoise)
- [How to test whitenoise config](https://whitenoise.readthedocs.io/en/stable/django.html#how-do-i-know-it-s-working)
- [Deploying whitenoise where django is not at the root of the domain](https://whitenoise.readthedocs.io/en/stable/django.html#deploying-an-application-which-is-not-at-the-root-of-the-domain)

**Parent Issue**: #219
[...]

PR #681, opened 2026-06-11
Create deployment docker setup (AWS series #671–675) #219

### LINKED ISSUE #219: Create deployment docker setup
### Overview

We need to add a separate deployment docker setup so that we can use it for automated deployment eventually. Differences are disabling debug variable, don't install development packages, etc.

### Action Items

- [ ] Create deployment Dockerfile and any necessary configs for production

### Resources/Instructions
- [production dockerfile in django dockerizing guide](https://testdriven.io/blog/dockerizing-django-with-postgres-gunicorn-and-nginx/#production-dockerfile)
- The task list is probably incomplete. Feel free to add to it. Things related to 12-factor apps would be good to do.
- [cookiecutter-django](https://github.com/cookiecutter/cookiecutter-django) is a template that generates boilerplate django projects. It might be a good reference for what a good deployment setup.

drakeredwind01:
Fixes #219

# AWS Deployment Series — Issues #671–675

A complete AWS production deployment setup built across five issues on branch `drake_670_integrate_whitenoise`.

## Summary

| Issue | Title | Key outcome |
|---|---|---|
| #671 | Split requirements into dev and AWS | Created `requirements-aws.in/txt`; corrected `psycopg2-binary` and added `whitenoise` |
| #672 | Add gunicorn and Whitenoise | Created `settings_aws.py`, `urls_aws.py`, updated `Dockerfile-aws` and `entrypoint-aws.sh` |
| #673 | Optimize Dockerfile-aws with multi-stage build | 18% image size reduction: 74MB → 60.7MB |
| #674 | Create docker-compose-aws.yml for local testing | Full local production test stack with docs |
| #675 | Update deployment workflow | Added `DOCKER_BUILDKIT=1`; updated `aws-resources.md` |

## Key discoveries across the series
- `psycopg2-binary` was incorrectly classified as dev-only in the issue — it's the database driver, required everywhere
- `DJANGO_ALLOWED_HOSTS` had no default, crashing `collectstatic` at Docker build time
- `drf_spectacular` turned out to be a runtime import dependency (used in `views.py` decorators), not dev-only — had to be added back to `requirements-aws.in`
- `requirements-aws.txt` compiled locally resolved packages incompatible with Python 3.10 in the Docker image — always compile with `--python-version 3.10`
- Named postgres volumes must differ between dev and aws compose stacks to avoid shared database state

---

# Split requirements into dev and AWS #671

## Response for HFLA re: requirements classification

`psycopg2-binary` is marked as a dev dependency in the issue, but it's the PostgreSQL database driver — the app cannot connect to the database without it in any environment. It should be included in `requirements-aws.in`.

`whitenoise` was not listed in the issue but has already been added to `requirements.in` on the current branch; it belongs in `requirements-aws.in` as well since it serves static files in production.

## Files created:
- app/requirements-aws.in — 9 production-only packages (excludes django-linear-migrations, drf-spectacular, markdown, and all pytest packages)
- app/requirements-aws.txt — compiled from the above; 16 resolved packages

## Files updated:
- docs/architecture/project_structure.md — added requirements-aws.in (10) and requirements-aws.txt (11) to the tree, renumbered setup.cfg to (12), and updated all caption descriptions to clarify dev vs. production purpose
- scripts/update-dependencies.sh — added a second uv pip compile command for the AWS requirements file so both stay in sync when dependencies are upgraded

---

# Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh #672

### 1. INSTALLED_APPS references dev-only packages not in requirements-aws.txt (serious)
`settings.py` includes `drf_spectacular` and `django_linear_migrations` in `INSTALLED_APPS`. Neither is installed in the production image, so the container will crash with `ModuleNotFoundError` on startup.
**Decision: create `app/peopledepot/settings_aws.py`** that imports from `settings.py` and overrides `INSTALLED_APPS` to exclude dev-only apps. This is the industry standard approach (used by Cookiecutter Django). Set `DJANGO_SETTINGS_MODULE=peopledepot.settings_aws` in `Dockerfile-aws`.
**Resolved: `app/peopledepot/settings_aws.py` created.**
Note: uses `from .settings import *` — the accepted Django idiom for settings inheritance (settings are a config namespace, not a module; explicit imports would require listing every setting). `# noqa: F401, F403` silences flake8, which doesn't know this is a settings file.

### 2. `collectstatic` during `docker build` crashes on missing `DJANGO_ALLOWED_HOSTS`
`settings.py:37`: `os.environ.get("DJANGO_ALLOWED_HOSTS").split(",")` raises `AttributeError` when the env var is unset at build time, preventing Django settings from loading at all.
**Decision: add a default** — `os.environ.get("DJANGO_ALLOWED_HOSTS", "localhost").split(",")` — so the build can proceed without that env var.
**Resolved: `settings.py:37` updated with default.**

### 3. `drf_spectacular` is a runtime import dependency — missed in #672, fixed in #674
`views.py` imports `drf_spectacular` at the module level for `@extend_schema` decorators. Removing it from `INSTALLED_APPS` in `settings_aws.py` wasn't enough — it also needs to be installed. It was incorrectly classified as dev-only; it's actually a runtime import dependency of the application code.
**Fix (applied in #674):** add `drf-spectacular` back to `requirements-aws.in`; keep it out of `INSTALLED_APPS` so schema endpoints and management commands stay inactive in production. Also created `urls_aws.py` to exclude the schema UI routes.

### 4. Dockerfile-aws has graphviz and Roboto font
These were copied from the dev Dockerfile and serve no production purpose. Leaving them in — out of scope for this issue, can be a separate issue if HFLA wants them removed.

## Files created:
- app/peopledepot/settings_aws.py — inherits base settings, overrides INSTALLED_APPS (removes drf_spectacular and django_linear_migrations) and REST_FRAMEWORK (removes DEFAULT_SCHEMA_CLASS)

## Files updated:
- app/requirements-aws.in — added gunicorn
- app/requirements-aws.txt — recompiled; now 18 packages including gunicorn 26.0.0
- app/peopledepot/settings.py — added "localhost" default to DJANGO_ALLOWED_HOSTS so collectstatic doesn't crash at build time
- app/Dockerfile-aws — uses requirements-aws.txt, sets DJANGO_SETTINGS_MODULE=peopledepot.settings_aws, runs collectstatic during build, adds BuildKit cache mounts, removes dev CMD
- app/entrypoint-aws.sh — runs migrate --noinput then starts gunicorn on 0.0.0.0:8000

---

# Optimize Dockerfile-aws with multi-stage build #673

### 1. Must build and measure the single-stage image before refactoring
The acceptance criteria requires a before/after size comparison. Need to build and record the current image size first, then implement the multi-stage build, then measure again. Do not start the refactor until the baseline is captured.

### 2. Size savings will be modest — I agree
The cache mounts (`--mount=type=cache`) added in #672 already keep pip/apk caches out of image layers, which was historically the biggest multi-stage win. Savings here will mainly be removing `pip` and `uv` from the final image. The issue itself flags this uncertainty; I confirmed the expectation going in.

### 3. `staticfiles/` must be explicitly copied from builder to runtime stage (foresight)
`collectstatic` runs in the builder stage and writes to `staticfiles/`. This directory must be explicitly `COPY --from=builder`'d into the runtime stage — whitenoise serves from it at runtime. Pre-emptively noted to avoid missing it during implementation.

## size before
docker build -f app/Dockerfile-aws -t peopledepot-aws:single-stage app/
docker image inspect peopledepot-aws:single-stage --format='{{.Size}}' | numfmt --to=inspect
    71M
docker images peopledepot-aws:single-stage
    IMAGE                          ID             DISK USAGE   CONTENT SIZE   EXTRA
    peopledepot-aws:single-stage   fded4b4e4602        330MB           74MB        

### reclaim space
docker rmi peopledepot-aws:single-stage

## size after
docker build -f app/Dockerfile-aws -t peopledepot-aws:multi-stage app/
docker images peopledepot-aws:multi-stage
    IMAGE                         ID             DISK USAGE   CONTENT SIZE   EXTRA
    peopledepot-aws:multi-stage   56eed5b30e55        282MB         60.7MB        

### reclaim space
docker rmi peopledepot-aws:multi-stage

## size diff
18% reduction (74MB → 60.7MB)

## Files updated:
- app/Dockerfile-aws — refactored to multi-stage; builder stage installs deps and runs collectstatic, runtime stage copies site-packages, gunicorn binary, and app code from builder; graphviz and font in runtime stage only

---

# Create docker-compose-aws.yml for local testing #674

### 1. Must NOT mount the app directory as a volume
`docker-compose.yml` (dev) mounts `./app/:/usr/src/app/` so developers can edit code and see changes instantly without rebuilding the image — this is intentional for development workflow.
`docker-compose-aws.yml` must NOT do this. The app code is baked into the image during the build (`COPY . .` in Dockerfile-aws). Mounting the directory would overwrite it with live local code, bypassing `collectstatic`, the multi-stage build, and `settings_aws.py` — nothing production-like would actually be tested.
**Proposed change:** omit the volume mount entirely from `docker-compose-aws.yml`.

### 2. Need a new `.env.docker-aws` file
`.env.docker` has `DEBUG=True` — wrong for production testing. `.env.prod.sample` has `SECURE_SSL_REDIRECT=True` — causes an infinite redirect loop locally since there's no SSL certificate. Neither works as-is.
**Proposed change:** create `app/.env.docker-aws` and `app/.env.docker-aws-example` based on `.env.prod.sample` but with `DEBUG=False`, SSL settings off, and local-friendly `DJANGO_ALLOWED_HOSTS`.

### 3. Port conflict with dev compose
Dev compose binds host ports `8000` (web) and `5432` (postgres). Running both simultaneously would fail.
**Proposed change:** use `8001:8000` and `5433:5432` in `docker-compose-aws.yml`. Dev keeps the familiar `localhost:8000`; production is accessible at `localhost:8001`. The compose file handles the mapping — developers don't need to remember the port numbers.
Port notation is `host:container`. The left side (host port) is where the conflict would happen — dev uses 5432, aws uses 5433, so both can run simultaneously. The right side (5432) is PostgreSQL's internal port inside each container — they're separate containers so that never conflicts.

### 4. Postgres named volume must differ from dev
Dev compose uses named volume `postgres_data`. If `docker-compose-aws.yml` used the same name, both environments would share the same database on disk — migration conflicts and data bleed between environments.
**Resolved: use `postgres_data_aws` in `docker-compose-aws.yml`.**

### 5. `.env.docker-aws` not covered by `.gitignore`
The gitignore lists explicit filenames rather than a wildcard, so `.env.docker-aws` would have been committed with secrets.
**Resolved: `.env.docker-aws` added to `.gitignore`.**

## Files created:
- `docker-compose-aws.yml` — builds Dockerfile-aws, ports 8001:8000 and 5433:5432, postgres_data_aws volume, no code volume mount
- `app/.env.docker-aws-example` — production-like env: DEBUG=False, SSL settings off, local ALLOWED_HOSTS; committed as template
- `app/.env.docker-aws` — gitignored developer copy of the above
- `docs/contributing/howto/test-aws-setup-locally.md` — full guide: setup steps, cleanup commands, comparison table of dev vs aws environments
- `docs/contributing/onboarding/test_aws_locally.md` — two commands and a link to the full guide

## Files updated:
- `.gitignore` — added `.env.docker-aws`
- `docs/contributing/howto/index.md` — added link to new guide

### drf_spectacular still imported via views.py
`views.py` imports `drf_spectacular` at the module level for `@extend_schema` decorators. Removing it from `INSTALLED_APPS` wasn't enough — `include("core.api.urls")` triggers `views.py` import at startup.
**Fix:** added `drf-spectacular` back to `requirements-aws.in` (it's a runtime import dependency, not dev-only); created `app/peopledepot/urls_aws.py` without schema UI routes; added `ROOT_URLCONF = "peopledepot.urls_aws"` to `settings_aws.py`.

### requirements-aws.txt compiled against wrong Python version
Recompiling locally resolved `rpds-py==2026.5.1` which requires Python>=3.11, but `Dockerfile-aws` uses Python 3.10 — build failed.
**Fix:** recompiled with `uv pip compile --python-version 3.10`; updated `scripts/update-dependencies.sh` to always pass `--python-version 3.10` for the aws compile command.

## Files created (including fixes):
- `docker-compose-aws.yml`
- `app/.env.docker-aws-example`
- `app/.env.docker-aws`
- `app/peopledepot/urls_aws.py` — production URL config without schema UI routes
- `docs/contributing/howto/test-aws-setup-locally.md`
- `docs/contributing/onboarding/test_aws_locally.md`

## Files updated (including fixes):
- `.gitignore` — added `.env.docker-aws`
- `app/requirements-aws.in` — added `drf-spectacular`
- `app/requirements-aws.txt` — recompiled targeting Python 3.10
- `app/peopledepot/settings_aws.py` — added `ROOT_URLCONF = "peopledepot.urls_aws"`
- `scripts/update-dependencies.sh` — added `--python-version 3.10` to aws compile command
- `docs/contributing/howto/index.md` — added link to new guide

---

# Update deployment workflow #675

## Most action items already handled pre-emptively in #671–674
- ✓ `deploy-dev.yml` already uses `docker build -f Dockerfile-aws`
- ✓ `requirements-aws.txt` is referenced inside `Dockerfile-aws` — picked up automatically
- ✓ `settings_aws.py` is set via `ENV DJANGO_SETTINGS_MODULE` in `Dockerfile-aws` — baked in
- ✓ `entrypoint-aws.sh` is the `ENTRYPOINT` in `Dockerfile-aws` — already correct
- ✓ gunicorn is started by `entrypoint-aws.sh` — already in place
- ✓ Workflow pushes to ECR and triggers ECS redeployment — unchanged and working

### 1. BuildKit not explicitly enabled — NOT mentioned in the issue
`Dockerfile-aws` uses `--mount=type=cache` which requires BuildKit. GitHub Actions enables BuildKit by default on recent Docker versions but it is not guaranteed and is not explicitly set in the workflow. If missing, the build fails with a syntax error on the `--mount` lines.
**Proposed change:** add `DOCKER_BUILDKIT=1` to the build step's `env` block in `deploy-dev.yml`.

### 2. Documentation update — IS mentioned in the issue
`docs/contributing/reference/aws-resources.md` already references `Dockerfile-aws` and `entrypoint-aws.sh` and is mostly current. Needs minor additions to mention gunicorn as the WSGI server, whitenoise for static files, and `settings_aws.py` as the production settings module.

## Files updated:
- `.github/workflows/deploy-dev.yml` — added `DOCKER_BUILDKIT=1` to build step env block
- `docs/contributing/reference/aws-resources.md` — step 5 updated to mention gunicorn and whitenoise; `settings_aws.py` added to Docker & App Startup section


fyliu (changes requested):
Thanks for working on this. But can you create separate PRs for each issue so that it's easier to review? These points may belong to different issues, but I can't be sure. There might be more issues, but these are the more visible ones.
- Please remove the first commit from the commit chain. That commit is attempting to implement a "not planned" issue.
- Please fix the code quality comment about importing all.
- Please test that the changes work before creating the PR. The MkDocs service is crashing.

PR #683, opened 2026-06-12
Add gunicorn and Whitenoise, and set up Dockerfile-aws and entrypoint-aws.sh #672

drakeredwind01:
Fixes #672

### Pre-implementation concerns

**1. Production requirements file needed**
`requirements.in` includes dev/test packages that shouldn't be installed in production. Created `requirements-aws.in` with only production-necessary packages. Note: `psycopg2-binary` is the PostgreSQL database driver and must be included — it is not dev-only. `whitenoise` is included since it serves static files in production.

**2. INSTALLED_APPS references dev-only packages not in requirements-aws.txt (serious)**
`settings.py` includes `drf_spectacular` and `django_linear_migrations` in `INSTALLED_APPS`. Neither is installed in the production image — container will crash with `ModuleNotFoundError` on startup.
**Decision:** create `settings_aws.py` that imports from `settings.py` and overrides `INSTALLED_APPS`. Set `DJANGO_SETTINGS_MODULE=peopledepot.settings_aws` in `Dockerfile-aws`.

**3. `collectstatic` during `docker build` crashes on missing `DJANGO_ALLOWED_HOSTS`**
`os.environ.get("DJANGO_ALLOWED_HOSTS").split(",")` raises `AttributeError` when unset at build time. Added `"localhost"` default.

**4. `drf_spectacular` is a runtime import dependency**
`views.py` imports `drf_spectacular` at the module level for `@extend_schema` decorators — removing it from `INSTALLED_APPS` was not enough. Added back to `requirements-aws.in`; kept out of `INSTALLED_APPS` so schema endpoints stay inactive in production. Created `urls_aws.py` to exclude schema UI routes and set `ROOT_URLCONF` in `settings_aws.py`.

**5. Dockerfile-aws has graphviz and Roboto font**
Copied from dev Dockerfile, no production purpose. Left in — out of scope, can be a separate issue.

### Files created:
- `app/requirements-aws.in` — production-only packages
- `app/requirements-aws.txt` — compiled targeting Python 3.10
- `app/peopledepot/settings_aws.py` — inherits base settings, overrides INSTALLED_APPS and REST_FRAMEWORK
- `app/peopledepot/urls_aws.py` — production URL config without schema UI routes

### Files updated:
- `docs/architecture/project_structure.md` — added requirements-aws files to tree
- `scripts/update-dependencies.sh` — added aws compile command with `--python-version 3.10`
- `app/requirements-aws.in` — added gunicorn, drf-spectacular
[...]

PR #684, opened 2026-06-12
Optimize Dockerfile-aws with multi-stage build #673

### LINKED ISSUE #219: Create deployment docker setup

### LINKED ISSUE #673: Optimize Dockerfile-aws with multi-stage build

drakeredwind01:
Fixes #673

### Pre-implementation concerns

**1. Size savings will be modest**
The cache mounts (`--mount=type=cache`) added in the previous PR already keep pip/apk caches out of image layers. Savings here are mainly removing `pip` and `uv` from the final image.

**2. `staticfiles/` must be explicitly copied from builder to runtime stage**
`collectstatic` runs in the builder stage and writes to `staticfiles/`. This directory must be `COPY --from=builder`'d into the runtime stage — whitenoise serves from it at runtime.

### Size comparison
- Single-stage: 74MB
- Multi-stage: 62.3MB
- **~16% reduction**

Note: slightly larger than original measurement due to `drf-spectacular` and its dependencies being correctly included in the production image (fixed in PR #683).

### Files updated:
- `app/Dockerfile-aws` — refactored to two stages: builder stage installs deps and runs collectstatic; runtime stage copies site-packages, gunicorn binary, and app code from builder

Closes #219


fyliu (changes requested):
@drakeredwind01 this PR will need to be looked at again since the `upstream/main` code changed and it doesn't rebase/merge anymore. It's really the last commit needing to be reapplied to `upstream/main`

PR #687, opened 2026-06-16
feat: create table UserPracticeAreaSecondaryXref (#680)

### LINKED ISSUE #680: Create Table: UserPracticeAreaSecondaryXref
### Overview

We need to create the UserPracticeAreaSecondaryXref table to support many-to-many relationship between UserPracticeAreaSecondary.

Used to be part of
- #601

### Action Items

- [ ] Create xref table UserPracticeAreaSecondaryXref
  field | type | related table | description
  --- | --- | --- | ---
  user | FK | User |
  practice_area | FK | PracticeArea |

- [ ] update models.py User model field 
  ```
    practice_area_secondary = models.ManyToManyField(
          "PracticeArea",
          related_name="users_secondary",
          blank=True,
          through="UserPracticeAreaSecondaryXref",
      )
  ```

### Resources/Instructions

- Reference similar code in models.py for guidance

egcuriel:
Fixes #680 

#### What changes did you make?
- **Models:** Created the UserPracticeAreaSecondaryXref model to serve as the custom through table for the practice_area_secondary many-to-many relationship.
- **API:** Updated `UserSerializer` to support the new secondary practice area structure and added an override to the serializer's `update` method to sanitize incoming payloads, silently discarding duplicate primary/secondary assignments before persisting the bridge records.
- **Admin:** Customized `UserAdmin` with a custom change form (`UserAdminForm`) to manage secondary practice areas via a virtual field. Implemented deferred M2M persistence by wrapping `save_m2m`, and added explicit form validation (`clean`) to actively block primary/secondary practice area conflicts.
- **Tests**: Added unit tests verifying string representation, persistence, unique compound constraints, and CASCADE deletion. Added integration tests verifying declarative relationship assignment (full set replacement) via PATCH. Asserts 200 OK on overlapping primary/secondary inputs and 400 Bad Request strictly on malformed data types.
- **Scripts:** Updated validate_mkdocs.sh to use modern Docker compose syntax.

#### Why did you make the changes? (we will use this info to test)
- Created the UserPracticeAreaSecondaryXref model to setup the necessary backend table that joins a user with multiple practice areas.
- To provide data integrity from Admin, custom validation was added in the Django Admin to block staff from saving duplicate primary/secondary records and displaying an error message to guide them to correct it.
- Implemented defensive data handling in the serializer to gracefully handle conflicting data and protect backend database integrity, preventing HTTP errors that could crash or break frontend applications.
- Updated existing model tests to ensure suite stability after the schema changes, and modernized infrastructure scripts to match current project standards.

#### Screenshots
<details>
<summary>Django Admin Validation - Demonstrates custom validation preventing duplicate primary/secondary selections in the Django Admin interface.</summary>
frontend_validation
</details>

<details>
<summary>API Form View - Verifies the DRF browsable API correctly renders the secondary practice area as a multi-select field.</summary>
api_serializer_mu
</details>

<details>
<summary>API JSON Response - Confirms the DRF serializer successfully returns a `200 OK` and correctlyexposes the new many-to-many relationship as a list.</summary>
api_update_mu
</details>


fyliu (changes requested):
Thanks for getting this done so quickly and the great clarification questions!

Now that I'm reading the changes, I realized my last comment in slack about calling it `practice_areas_secondary` was confusing. I meant you should name the User model field `practice_areas_secondary`. You were correct before to name the relate_name `"users_secondary"`. Can you change it back? Thanks!

fyliu (changes requested):
Thanks!

Looks like there's a test failure. Maybe the it needed updating.

<details>
<summary>Details</summary>

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "uniq_global_user_check"
DETAIL:  Key (user_id, check_type_id)=(7ed5ee45-9d34-4091-bb8e-8cba1e28f02f, ed02c501-12cc-44b4-bf6c-40c846a5d9c1) already exists.
=========================== short test summary info ============================
FAILED core/tests/test_api.py::test_api_prevent_duplicate_global_usercheck - ...
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
!!!!!!!!!!!! xdist.dsession.Interrupted: stopping after 1 failures !!!!!!!!!!!!!
======================== 1 failed, 42 passed in 21.59s =========================
```

</details>

fyliu:
Nevermind the above comment. I ran update-requirements locally, and the newer version is what broke the test.

egcuriel (commented):
- Models: Implemented UserPracticeAreaSecondaryXref model to serve as a custom through-table, adding full_clean() validation to act as a database-level failsafe against duplicate primary/secondary practice area assignments.

- Admin: Refined the ValidationError copy in admin.py to provide clear, actionable feedback to staff if they attempt to save conflicting data.  

- Naming Conventions: Updated the User model fields and related names to practice_areas_secondary and users_secondary per previous review feedback.  

- Verification: All automated tests are passing, and the integrity of the data is now protected at both the form and model layers


I went in to double check my code didn't break anything and realized a validation check in my model couldn't hurt the database

egcuriel:
Rebased onto upstream/main, resolved the migration file conflict, and pushed a clean 4-commit history (including the separate validate_mkdocs.sh fix as discussed). Ready for review!

app/core/api/serializers.py
```diff
@@ -128,6 +133,49 @@ class Meta:
             "email",
         )
 
+    def update(self, instance, validated_data):
```
fyliu, line 172: There's a discrepancy between silently dropping the duplicate here and the model's raising a validation error. It's better to return an error from the serializer as well, so the API client knows there's a problem with the input. The validation logic would be in `validate()` of course.

app/core/api/serializers.py
```diff
@@ -128,6 +133,49 @@ class Meta:
...
+
+            # Override practice_area_primary_id if a new practice_area_primary is included in this payload.
+            if "practice_area_primary" in validated_data:
+                primary_incoming = validated_data["practice_area_primary"]
+                practice_area_primary_id = getattr(
+                    primary_incoming, "pk", primary_incoming
+                )
+
+            # Delete the existing secondary practice_area records for this user.
+            UserPracticeAreaSecondaryXref.objects.filter(user=instance).delete()
```
fyliu, line 182: Maybe mention somewhere that this is doing a delete and recreate strategy and the `created_at` timestamps and original UUIDs are lost.

fyliu (changes requested):

fyliu (changes requested):
Looks good to me. But I clicked the copilot button to see if it has any suggestions and it said to:
- enclose the delete and create in a transaction
	- in case something happens between the delete and create
- deduplicate the practice_areas list
	- we don't really allow it, but it might still happen in unexpected cases
It makes sense to me, so I'm pasting the code here to see if it makes sense to you too.
```py
from django.db import transaction


def update(self, instance, validated_data):
    secondary_areas = validated_data.pop(
        "practice_area_secondary",
        None,
    )

    with transaction.atomic():
        # Update the User first. If this fails, existing xrefs remain untouched.
        instance = super().update(instance, validated_data)

        # None means the field was omitted from a PATCH.
        # An empty list means explicitly clear the relationship.
        if secondary_areas is not None:
            # PrimaryKeyRelatedField returns PracticeArea instances.
            # dict.fromkeys() removes duplicates while preserving request order.
            secondary_ids = list(
                dict.fromkeys(area.pk for area in secondary_areas)
            )

            UserPracticeAreaSecondaryXref.objects.filter(
                user=instance
            ).delete()

            UserPracticeAreaSecondaryXref.objects.bulk_create(
                [
                    UserPracticeAreaSecondaryXref(
                        user=instance,
                        practice_area_id=practice_area_id,
                    )
                    for practice_area_id in secondary_ids
                ]
            )

    return instance
```

egcuriel:
Thank you for the feedback @fyliu ! I've updated the following:

1. Wrapped the entire update and Xref creation in a `transaction.atomic()` block to prevent any data corruption if a step fails.
2. Added the `dict.fromkeys()` deduplication to handle duplicate IDs while preserving the exact order submitted by the frontend, just in case we need to support ranking later.
3. Added `ignore_conflicts=True` to the `bulk_create()` call as a final defensive layer against the `UniqueConstraint` errors on the model.
4. Added a new test in `test_api.py` to explicitly verify that the deduplication logic handles duplicate payloads successfully without throwing a 500 error.

I also adapted the variable names slightly (e.g., using `practice_area_secondary` instead of `secondary_areas`) so they stay perfectly consistent with the model and the payload fields I set up.

Pushed those changes up, let me know if you need anything else!

